PR #137, opened 2011-03-14
Case-insensitivity for TaggedModel.tagged_with for PostgreSQL database

ghost:
Hi All, 

For our own project we needed the tagged_with model to be case insensitive for PostgreSQL because the project is running on Heroku.

I've added a like_operator method, I'm using it in tagged_with and added two isolated unit tests to confirm that it is working.

Hopefully this is usefull for the original development track as well.

All the best,
Jevado / Presentplus


StarWar:
I need case sensitivity while searching, I'm getting same records for distinct tags with case sensitive different names.

Issue #464, opened 2014-01-21
another problem with multibyte tags


nir0:
gem version is 3.0.1
error: Tag cannot be empty

```
2.0.0p195 :001 > post = NewsPost.new(id: 277423, 
2.0.0p195 :002 >           name: "В торговом центре 52-летний мужчина в туалете свел счеты с жизнью", 
2.0.0p195 :003 >           markdown: "В Тюмени проводится проверка по факту обнаружения в одном из торговых центров тела мужчины. Как сообщили \"Вслух.ру\" в пресс-службе регионального управления СК РФ, накануне, 19 января, около 21.30 в закрытой туалетной кабине на первом этаже в одном из торговых центров Тюмени обнаружено тело 52-летнего мужчины, покончившего жизнь самоубийством - воткнувшего нож себе в грудь. Проводится доследственная проверка.
2.0.0p195 :004"> 
2.0.0p195 :005"> Установлено, что дверь кабины была заперта изнутри. О случившемся сообщили посетители торгового центра, которые в тот момент находились в этом же помещении и услышали странный хрип. Следствие проводит комплекс необходимых мероприятий, направленных на установление всех обстоятельств произошедшего. Опрашиваются очевидцы и родственники погибшего. По результатам проверки будет принято решение.", 
2.0.0p195 :006 >           text: "<p>В Тюмени проводится проверка по&#160;факту обнаружения в&#160;одном из&#160;торговых центров тела мужчины. Как&#160;сообщили &#171;Вслух.ру&#187; в&#160;<span class=\"nobr\">пресс-службе</span> регионального управления СК&#160;РФ, накануне, 19&#160;января, около 21.30 в&#160;закрытой туалетной кабине на&#160;первом этаже в&#160;одном из&#160;торговых центров Тюмени обнаружено тело <span class=\"nobr\">52-летнего</span> мужчины, покончившего жизнь самоубийством &#8211; воткнувшего нож&#160;себе в&#160;грудь. Проводится доследственная проверка.</p>
2.0.0p195 :007"> 
[...]

PR #498, opened 2014-03-19
Use database's lower function for case-insensitive match

leo-souza:
This PR is for the same problem as PR #492 but with a diferent solution, since it was breaking the build. 
It's similar to PR #472 and relates to issue #464


sintro:
https://github.com/mbleigh/acts-as-taggable-on/issues/464#issuecomment-214843950

sintro:
As I described here https://github.com/mbleigh/acts-as-taggable-on/issues/464#issuecomment-214843950, the problem is still exists, and it depends on the database adapter and locale settings. In that case, all the code, which includes databases lower() functions will not work for Postgres, if the database uses LC_COLLATE locale not corresponding to tag`s language. It is in many cases more handy to have the LC_COLLATE using "C" locale.
May be, this gem need refactoring all the database`s case-functions to Ruby`s analogues?

PR #578, opened 2014-08-20
touch tagging belongs_to associations

zshannon:
It'd be helpful for caching if taggable objects were touched when their tags change.


stevenbarragan:
:+1: for the idea!

dedman:
Any update with this? Would be a useful feature to have.

Loremaster:
This is still isn't merged? 😢

PR #598, opened 2014-10-05
Added namespacing functionality

JakeTheSnake3p0:
Description of pull request: To allow for namespacing of the database tables. It is useful for situations where a mounted engine and the app itself both utilize acts-as-taggable-on but want to keep the taggings tables separated.

I am not a smart man when it comes to git merging/rebasing, so I apologize if it looks like a messy pull request. All tests are passing, though keep in mind I had to modify the test suite to actually test the namespaced models alongside the regular ones. I don't believe I changed the _contents_ of the tests though.

I also noticed there was another pull that came in before this one; not sure how that'll play out in this pull.

I have tested this out in a new basic app and it's working.

### DIFF
--- a/lib/acts_as_taggable_on/taggable/core.rb
+++ b/lib/acts_as_taggable_on/taggable/core.rb
@@ -112,14 +112,14 @@ module ActsAsTaggableOn::Taggable
             tags_conditions = tag_list.map { |t| sanitize_sql(["#{namespaced_class(:Tag).table_name}.name #{ActsAsTaggableOn::Utils.like_operator} ?", t]) }.join(' OR ')
           end
 
-          conditions << "#{table_name}.#{primary_key} NOT IN (SELECT #{namespaced_class(:Tagging).table_name}.taggable_id FROM #{namespaced_class(:Tagging).table_name} JOIN #{namespaced_class(:Tag).table_name} ON #{namespaced_class(:Tagging).table_name}.#{namespaced(:tag_id)} = #{namespaced_class(:Tag).table_name}.#{namespaced_class(:Tag).primary_key} AND (#{tags_conditions}) WHERE #{namespaced_class(:Tagging).table_name}.taggable_type = #{quote_value(base_class.name, nil)})"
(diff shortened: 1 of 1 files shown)


fletcher91:
Does this mean rows in the tag table are tied to scope? E.g. a tag 'foo' in the context 'bar' is different from the tag 'foo' in the context 'baz' in the tags table (aka it having a different ID)

Some background; we need tags to be scoped/namespaced to a forum, and since some aren't public, leaking tags across them (with autocomplete) is a no-no.

This is a blocking issue in my project, so let me know if I can help.

JakeTheSnake3p0:
From what I recall when writing, namespacing only applied to database table names, not the data that was stored in those tables.

@fletcher91, would it not be possible to utilize "Dynamic Tag Contexts" in this case? Something like the following:

```ruby
@forum.set_tag_list_on("forum_#{ @forum.slug }", "some, tags")
```

...and then in the logic of your autocomplete, retrieve all tags within the context of `forum_*`? Just spit-balling, I'm not sure if it will work.

fkoessler:
Does that mean we can't use AATO when working with namespaced models until this pull request is merged? @JakeTheSnake3p0 @seuros

seuros:
@fletcher91 you can. that feature is to namespace the tags/taggings tables

JakeTheSnake3p0:
@fkoessler, this pull request is specifically for adding a prefix to the AATO tables in the database.

So originally the table would be called `taggings`, but with a namespace prefix of `manager`, the table name would result in `manager_taggings`. The major benefit of this is for example in engines/gems. The engine/gem can have its own taggings table and the main application which utilizes the engine can have its own as well.

A side benefit of namespacing the tables for me is database synchronization between applications; I synchronize all tables prefixed with `manager_`. This way I can synchronize taggings between applications and also let each application have its own taggings as well.

fkoessler:
@JakeTheSnake3p0  Understood ;) Great job.

derekcannon:
Any idea when this feature is scheduled for release? I am trying to get acts_as_taggable_on working on a project that already has a Tags table. From what I can tell, there is no way to change the expected table from Tags to something else. This feature seems to solve that issue.

m5rk:
Is this PR abandoned? @JakeTheSnake3p0 as the author, perhaps you can provide a little more insight into the motivation for this PR in this PR's description? Also, update the README as appropriate?

JakeTheSnake3p0:
This PR was to be merged/utilized in the 4.0 release, but I think it must've gone off the radar. I haven't kept up to date with the master branch so I don't know what conflicts this will create at this point in time. My README has already been updated to reflect the purpose of this pull request but again, it is out of sync with the main branch's README.

I don't think this will solve your situation m5rk. All this merge does, is add a user-defined prefix to the AATO database tables and tweak the surrounding code to make writing/reading work as expected.

PR #602, opened 2014-10-20
sha_prefix should not be random

jdelStrother:
Hi there - `Utils::sha_prefix(string)` essentially returns a completely random hash every time you call it.  This is unexpected (at least to me), and prevents query caching where it's used as the join table alias.

I think the random behaviour originated in 713eb2919a4e63052d85e2aa9dfc4a59480514e4 (where it started out as a unique identifier), and got subsequently more confusing with be398de9ac6c9b801fdcb7f51ffe136fd2f68a42 and d9f0b77c09c5f9aab6099fc4c8d18326de58e842

How about this change?

### DIFF
--- a/lib/acts_as_taggable_on/utils.rb
+++ b/lib/acts_as_taggable_on/utils.rb
@@ -18,7 +18,7 @@ module ActsAsTaggableOn
       end
 
       def sha_prefix(string)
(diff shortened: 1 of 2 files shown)


idoo:
now it fails when you're combine queries and have same tag name for both

Issue #227, opened 2012-02-03
Mysql2::Error: Duplicate entry 'şi' for key 'index_tags_on_name' on diacritic letters


infoman:
I have a tags table in my database with tag name column marked as unique index (as I do not want to have two identical tags).

When I run a script importing multiple different objects with many different tags there's an issue that mysql counts the diacritic and non-diacritic versions of one letter (for example, s and ş) the same, but this gem still tries to add a new tag (even when "select \* from tags where name = 'şi'" returns one record with tag 'si').

The same error also occurs when providing an array with non-unique records as tags_list.

Possible fix can be to catch ActiveRecord::RecordNotUnique exception and then use the tag returned by select.

1v:
> rake acts_as_taggable_on_engine:tag_names:collate_bin

Not working.

1v:
I ended up just commenting line that cause error:
```ruby
AddMissingUniqueIndices.class_eval do
  def self.up
    # add_index :tags, :name, unique: true

    remove_index :taggings, :tag_id if index_exists?(:taggings, :tag_id)
    remove_index :taggings, [:taggable_id, :taggable_type, :context]
    add_index :taggings,
              [:tag_id, :taggable_id, :taggable_type, :context, :tagger_id, :tagger_type],
              unique: true, name: 'taggings_idx'
  end

  def self.down
    # remove_index :tags, :name

    remove_index :taggings, name: 'taggings_idx'

    add_index :taggings, :tag_id unless index_exists?(:taggings, :tag_id)
    add_index :taggings, [:taggable_id, :taggable_type, :context]
  end
end
```

seuros:
This was fixed in a later release. If it still happens on v13 with Rails 8, please open a new issue with a repro.

PR #604, opened 2014-10-24
Fix encoding/accent raising  ActsAsTaggableOn::DuplicateTagError 

randoum:
Creating a tag named "spécifique", then a tag named "specifique" would raise ActsAsTaggableOn::DuplicateTagError

SQL server being correctly configured recognizes that `"spécifique" == "specifique"`, but acts-as-taggable-on does not.

---

I implemented 2 things in this PR, sorry for the confusion but I was obliged to fix that spec bug in order to correctly deliver my PR. Here are the details:

**1) fix rspec bug**

Several specs was setting `ActsAsTaggableOn.strict_case_match = true` without reinitializing it after, which may input invalid settings for the other specs, and all depending on the order specs are run. I cleaned that by replacing all instances of
```
it 'should do something' do
  ActsAsTaggableOn.strict_case_match = true
  ...
end
```
by
```
it 'should do something', strict_case_match: true do
  ...
end
```

The `strict_case_match` flag correctly sets `ActsAsTaggableOn.strict_case_match = true` before the examples and reset it to `false` after the examples. 

**2) Use I18n transliterate for unicode comparison**

The main reason of this PR: In order to correctly compare strings such as "spécifique" and "specifique" I added `I18n.transliterate` to `Tag#comparable_name`. This also has the advantage to support I18n, so users of acts-as-taggable-on can use I18n to precisely define unicode comparison according to their language.

Example:
In english "Jürgen" would be compared to ASCII "Jurgen"
In german "Jürgen" would be compared to ASCII "Juergen"
And this can be set in locale yml files

More info: http://api.rubyonrails.org/classes/ActiveSupport/Inflector.html#method-i-transliterate

---

Seems to be also a fix for #227 
Let me know what you guys think

### DIFF
--- a/lib/acts_as_taggable_on/tag.rb
+++ b/lib/acts_as_taggable_on/tag.rb
@@ -105,7 +105,7 @@ module ActsAsTaggableOn
         if ActsAsTaggableOn.strict_case_match
           str
         else
-          unicode_downcase(str.to_s)
+          I18n.transliterate(unicode_downcase(str.to_s))
         end
       end
 
--- a/spec/acts_as_taggable_on/tag_spec.rb
+++ b/spec/acts_as_taggable_on/tag_spec.rb
@@ -115,8 +115,7 @@ describe ActsAsTaggableOn::Tag do
         include_context 'without unique index'
       end
 
-      it 'should find by name case sensitive' do
-        ActsAsTaggableOn.strict_case_match = true
+      it 'should find by name case sensitive', strict_case_match: true do
         expect {
           ActsAsTaggableOn::Tag.find_or_create_all_with_like_by_name('AWESOME')
         }.to change(ActsAsTaggableOn::Tag, :count).by(1)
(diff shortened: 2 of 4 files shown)


joost:
+1 for this fix

ahmetabdi:
:+1:

lib/acts_as_taggable_on/tag.rb
```diff
@@ -105,7 +105,7 @@ def comparable_name(str)
         if ActsAsTaggableOn.strict_case_match
           str
         else
-          unicode_downcase(str.to_s)
+          I18n.transliterate(unicode_downcase(str.to_s))
```
seuros: what about existing data ? I think this is a breaking change.

nburt:
I've also run into this, my tag list has "montreal" and "montréal". I'm also not able to remove "montréal" from the tag list using tag_list.remove so we've been doing it manually with SQL queries.

Has anyone come up with a fix/workaround for this?

randoum:
@seuros as a collaborator, can you advise/help ?
Thanks in advance

rikettsie:
Hi @nburt, are you using MySql? If so, you can enable binary collation.
Via a rake task:
rake acts_as_taggable_on_engine:tag_names:collate_bin
Or via an initializer:
ActsAsTaggableOn.force_binary_collation = true

nburt:
Thanks @rikettsie.
We are using MySql, I'll check this out tomorrow morning and let you know if that fixes it.

PR #622, opened 2015-01-06
clears column cache on reset_column_information resolves #621

### LINKED ISSUE #621: @acts_as_taggable_on_cache_columns should be reset by reset_column_information
If its not reset then columns will continue to return the cached value and not requery the database as expected.

warp:


seuros:
LGTM

seuros:
Thank you.
Will release it shortly.

Issue #623, opened 2015-01-12
Duplicate entries for tags with spÉcial characters


bbnnt:
Updated and all, but the problems remains.

If for example the word "Numérique" is saved, and if it already exists, acts_as_taggable breaks.
Same will go for "numerique".

Is there a way to handle it ?

seuros:
You have to either provide a test case or backtrace.
If you are using Mysql, check your encoding.

bbnnt:
Sorry for that 

Here is a gist of what occurs : https://gist.github.com/bbnnt/7aadc59bea0edb24630d
"Numerique" does not exists nor saves. It is actually because of the key, as far as I can understand, but I really don"t know how to solve it

Encoding of the table is UTF-8 unicode and collation is utf-8_general_ci

seuros:
I can't reproduce this bug.

seuros:
Can you reproduce it with a fresh database/app ?

bbnnt:
Yes :/
I've done a rake db:drop db:create db:migrate

Then tried again.
https://gist.github.com/bbnnt/d5a5489182ac7bfe9be8
( I've highlighted the commands to make it clearer here http://cl.ly/image/1F391b0n3O1x?_ga=1.72731347.1360774124.1418737841 )

inket:
Could reproduce this bug. It also happens with the Japanese space -> "　"

bbnnt:
Too bad it is not such an issue for the maintainer
it sure works when you start with an empty db, but it gets awful if you start to import existing data, for some very obscure and unobvious reasons

seuros:
@bbnnt  : `Too bad it is not such an issue for the maintainer ad the logs provided pretty much shows that it does occurs` 
That not nice from you, this is an OpenSource project you can step up, fix it and send a PR. 
Your argument will be valid if i refused/ignored to merge a fix.

bbnnt:
@seuros that was not an arguent; but more or less to get an reply from you (:
I'll get into this

seuros:
:+1:

bf4:
@bbnnt This project is really hard to maintain.  It's a really old code base with lots of bugs, frequently reported issues, and none of the maintainers, as far as I know, use it anymore.

bbnnt:
@bf4 dam i'm getting old-schooled ! would you recommend another gem that has similar functionalities ?

seuros:
[...]

PR #633, opened 2015-02-26
#623: updated README (added suggestion for manual column alteration).

rikettsie:
This is to fix special characters used in tag names (MySql only).
A tag 'name' is stored by the gem as 'binary encoded string', but if collation is not specified as 'utf8_bin' for that column, all the comparisons are not made properly and thus the unicity constraint expressed by the index 'index_tags_on_name' generate the error you experienced.

Thus for a quick circumvention, you could alter the 'tags' table in the column 'name':
<pre>
ALTER TABLE tags MODIFY name VARCHAR(255) CHARACTER SET utf8 COLLATE utf8_bin;
<pre>

### DIFF
--- a/README.md
+++ b/README.md
@@ -57,9 +57,14 @@ Review the generated migrations then migrate :
 rake db:migrate
 ```
 
-MySql users should also apply a column change to get special characters
-work correctly for tag names, see [issue #623](https://github.com/mbleigh/acts-as-taggable-on/issues/623).
-Execute the following command in the MySql console:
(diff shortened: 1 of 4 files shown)


seuros:
This is awesome. :green_heart:

rikettsie:
Yay! :)
I've just pushed a customized migration for MySql and a rake task in my current master. I would also like to add a specific test.

seuros:
Go for it :+1:

PR #652, opened 2015-05-08
Rails 5 Compatibility

chrisb:
The util method `active_record4?` does not account for major versions higher than `4` (like `5`).

This PR patches association/relation support for those of us on edge Rails.


seuros:
There is much more change than this patch. I will do it when Rails 5 alpha is out, meanwhile the API might change.

chrisb:
No problem. Looking forward to a Rails 5 branch soon, then. Cheers! :beers:

PR #653, opened 2015-05-09
Use sanitize instead of quote_value

ianfleeton:
In Rails master quote_value now only accepts 1 param and aliases
sanitize.

sanitize has existed for a while only accepting 1 parameter
and delegates to connection.quote just as quote_value did so this should
preserve current behaviour (column name was never provided, only nil)
and help future compatibility.


ianfleeton:
Issue addressed in d7ac3b0

PR #654, opened 2015-05-09
Active Record 5 also requires a scope parameter for has_many

ianfleeton:


ianfleeton:
Issue addressed in d7ac3b0

PR #673, opened 2015-08-01
Add rails 5.0.0 alpha support

FlowerWrong:
Nowadays, I am using with rails 5 alpha. And I like [acts-as-taggable-on](https://github.com/mbleigh/acts-as-taggable-on). After use it, I found it was not suport rails  5, So I add the suport for it.
The main problem are `count` and `quote_value` methods of activerecord.
See https://github.com/rails/rails/blob/master/activerecord/lib/active_record/sanitization.rb#L10 and https://github.com/rails/rails/blob/master/activerecord/lib/active_record/relation/calculations.rb#L38 .
I just add support for them, like this
```ruby
def count(column_name = :all, options = {})
  ActsAsTaggableOn::Utils.active_record5? ? super(column_name) : super(column_name, options)
end
```
:8ball:


seuros:
We should drop support for rails 3.2 then, that will clean up the codebase a little bit.

PR #697, opened 2015-11-05
scope tags to specific tagging context

zapnap:
I found this to be pretty valuable (no need to subclass for my use case):

```ruby
ActsAsTaggableOn::Tag.for_context('skills') => # tags applied to 'skills' context, but not 'needs' context
```

### DIFF
--- a/lib/acts_as_taggable_on/tag.rb
+++ b/lib/acts_as_taggable_on/tag.rb
@@ -50,6 +50,12 @@ module ActsAsTaggableOn
       where(clause)
     end
 
+    def self.for_context(context)
+      joins(:taggings).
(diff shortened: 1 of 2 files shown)


snoblenet:
I am using acts-as-taggable-on 3.5.0. I have: 

````
class Report < ActiveRecord::Base
  acts_as_taggable_on :categories, :countries
end
````

Why do I get this?

````
> ActsAsTaggableOn::Tag.for_context("categories").most_used(10)
NoMethodError: undefined method `for_context' for #<Class:0x007fd6303d30c0>
````

Many thanks,

Steven

andrezimpel:
+1 @snoblenet

PR #702, opened 2015-11-15
Remove support for Rails 3.2

bram70:
Remove support for Rails 3.2


lib/acts_as_taggable_on/compatibility.rb
```diff
@@ -3,8 +3,6 @@ def has_many_with_taggable_compatibility(name, options = {}, &extention)
     if ActsAsTaggableOn::Utils.active_record4? || ActsAsTaggableOn::Utils.active_record5?
```
seuros: we won't need this  since we dropped support of 3.2

PR #706, opened 2015-11-19
README.md, Improved documentation for Tag Ownership

schnmudgal:
Documentation for adding and removing Owned tags with example.

There have been many issues for the case when **Tags are added with owner, it deletes previous tags**.
Updated documentation to handle them(with example).


seuros:
Thank you 👍

PR #709, opened 2015-12-20
Revert "Added missed indexes."

fabn:
Reverts mbleigh/acts-as-taggable-on#682

I'm investigating some performance issues and I found out this PR while looking into this repo.

I think this PR should be reverted because it creates some useless indexes and the other created indexes will likely never be used and only increases tag creation times.

By going into details

``` ruby
add_index :taggings, :tag_id
```

This is already provided by [this line](https://github.com/mbleigh/acts-as-taggable-on/blob/master/db/migrate/2_add_missing_unique_indices.rb#L7) as leftmost prefix of an existing index

``` ruby
add_index :taggings, :taggable_id
```

This is redundant because is a left prefix of the index below and also already provided by [this index](https://github.com/mbleigh/acts-as-taggable-on/blob/master/db/migrate/4_add_missing_taggable_index.rb#L3)

``` ruby
add_index :taggings, :taggable_type
add_index :taggings, :context
```

These two index as is may provide some benefits if one is doing some sort of stats (e.g. tags vs keywords, etc) but I don't think they should be added by default.

``` ruby
add_index :taggings, :tagger_id 
```

Left prefix of below index

``` ruby
add_index :taggings, [:tagger_id, :tagger_type]
```

This is the only truly missing index and can be useful if one wants to show user tagged elements.

``` ruby
add_index :taggings, [:taggable_id, :taggable_type, :tagger_id, :context], name: 'taggings_idy'
```

This is already provided by [this migration](https://github.com/mbleigh/acts-as-taggable-on/blob/master/db/migrate/4_add_missing_taggable_index.rb#L3), it only add additional columns. If this is going to be added then the other should be removed in favor of this.

In conclusion in my opinion #682 as is should be reverted. Maybe a good replacement would only include 
a commented index for `taggable_type` and `context` explaining that it can be useful only to do some stats and a commented version of `add_index :taggings, [:tagger_id, :tagger_type]` that explain its usage.

In this way upgrading users will not add them blindly only because it's description says "Add missing indexes".

What do you think?


CSammy:
I have been working on quite large databases for a while now and I agree with @fabn on every point he made. The only thing I'm undecided on is his suggestion to provide deactivated indices with comments on when to activate them.

denschub:
/cc @mbleigh @seuros 

We (as in the diaspora\* proejct) use quite a heavy tag base and these indexes are causing nothing but additional database load by adding indexes that won't get used. What's the state of this PR?

razum2um:
Doubled everyone above, besides indexes on single columns like `taggable_type` and `context` have low selectivity anyway and only slow things down

seuros:
Let get a new release out this week.

denschub:
Thank you.

PR #721, opened 2016-01-11
Update clean! method to use case insensitive uniq! when strict_case_m…

mark-jacobs:
…atch is false

The clean! method in tag_list.rb uses uniq! to remove duplicates from the list. When strict_case_match is false it will not remove duplicates that are the same words but differently cased.

I've changed this so that these duplicates will still be removed, so a tag_list containing ["one","One","OnE"] will become ["one"] when strict_case_match is false.


chrismhilton:
:+1:

Issue #342, opened 2013-04-15
Form with nested taggable models don't update *_list attributes


shekibobo:
I'm seeing this problem in an app I'm working on. I have the following setup:

``` ruby
class Parent < ActiveRecord::Base
  has_many :children
  accepts_nested_attributes_for :children, allow_destroy: true
  attr_accessible :children_attributes
end

class Child < ActiveRecord::Base
  belongs_to :parent
  acts_as_taggable_on :categories, :filters, :keywords
  attr_accessible :title, :category_list, :filter_list, :keyword_list
end
```

Now, I have a form for `Parent` which has fields for `Child` which include text inputs for `category_list`, `filter_list`, and `keyword_list`.

If I have a form specifically for `Child` with these fields, I can create and edit them just fine.

In the nested form, however, I can create them, but editing them doesn't ever work.

At first I thought it was related to the dirty module on multiple contexts, but fixing that doesn't change anything. I finally wrote some feature specs to test out the behavior, and I'm finding that even if I change something that isn't from `acts_as_taggable_on` changes with the `list` fields, ensuring that the child object would be dirty and thus updated, the tests still fail. The `title` attribute changes successfully, but the `*_list` attributes will not change from a nested form.

spodlecki:
I just ran into this as well... did you find a fix? From what I can tell, the Child is not being marked as being dirty... I tried to mess around with the gem, but found this was just as simple...

```
class Bread < ActiveRecord::Base
  acts_as_taggable
  acts_as_taggable_on :descriptors

  def descriptor_list=(values)
    set_values = super(values)
    updated_at_will_change! if descriptor_list_changed?
    set_values
  end
end
```
You can change the `update_at` to say, an `id` or something else... I just used updated_at because nearly all of our records use a timestamp of some nature.

micahlisonbee:
`descriptor_list_changed?` is always false for me in my case. Any tips?

spodlecki:
I haven't ran into the descriptor_list not being marked dirty, are you sure it is being set? That seems odd to me that it wouldn't be working.

We currently use #723 in our gemfile which includes the fix. I'm not sure where acts-as-taggable-on stands in their master branch.

toobulkeh:
Problem still persists today. Please merge #723.

seuros:
Believed fixed on current versions. If you can still reproduce it on v13, open a fresh issue and we will take a look.

PR #723, opened 2016-01-21
Cached Attributes (Rails 4) & marking nested models as dirty

spodlecki:
https://github.com/mbleigh/acts-as-taggable-on/issues/342
- Clear local column cache when parent does
- Get the model to be marked as dirty when setting a new list via a context

Model

``` ruby
class TaggableModel < ActiveRecord::Base
  acts_as_taggable
  acts_as_taggable_on :languages
end
```

Controller

``` ruby
class FoosController < ContentAdminController
      # .....
      def update
        if @foo.update_attributes(update_params)
          redirect_to foo_path(@foo), flash: {success: 'Updated foo.'}
        else
          render :edit
        end
      end
      # .....
    private

      def update_params
        params.require(:foo).permit(
          :taggable_attributes => [
            :id, :_destroy, {:language_list => []}
          ]
        )
      end
end
```

Post Data looks like...

``` json
{
  "utf8"=>"✓",
  "authenticity_token"=>"some-tada-here",
  "foo"=>{
    "taggable_attributes"=>{
      "0"=>{
        "_destroy"=>"0",
        "language_list"=>"english, french, german",
        "id"=>"15"
      }
    }
  },
  "commit"=>"Save",
  "id"=>"1"
}
```

Given the above, the TaggableModel will not receive a Dirty state, which ends up with an unsaved child.

### DIFF
--- a/lib/acts_as_taggable_on/taggable/cache.rb
+++ b/lib/acts_as_taggable_on/taggable/cache.rb
@@ -40,10 +40,25 @@ module ActsAsTaggableOn::Taggable
           end
         end
 
+        ###
+        # Clear local column cache when parent does
+        # https://github.com/jpregracke/acts-as-taggable-on/commit/92ef120b9716e392285c1617f2e6838b8da45e00
+        #
         def reset_column_information
           super
+          clear_acts_as_taggable_on_cache_columns
+        end
+
+        def clear_caches_calculated_from_columns
+          super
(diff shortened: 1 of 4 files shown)


Erowlin:
Hey, thanks for the fix @spodlecki ! 

Also, it looks similar to #728 and #812. 

Thanks a lot !

Issue #257, opened 2012-05-18
*_list attribute not getting dirtied when updated has nested attribute


budu:
When you call `update_attrbutes` on a model which have nested attributes for a model where the only attribute that changes is a `*_list`, it doesn't work as the nested record doesn't get dirty.

Skulli:
I am stil having issues with this on (version 4.0.0 and rails 4.2.7)

*empty start*

    g.document.marker_list = "test1, test2"
    g.save

that works, g.document.dirty? is also true
next

     g.document.marker_list = "test"
     g.save

that wont, also adding another one. It only works on the initial assignment when adding tags the first time. Document is not dirty.

to make it work i have to manually call

    g.document.save

PR #728, opened 2016-02-07
Bugfix dirty attributes handling for `tag_list=`.

lukeasrodgers:
`tag_list` was not being correctly marked as dirty if it was modified on a preexisting record.
This was caused by two factors:
- Inconsistent use of singular/vs plural tagging contexts.
- `#custom_contexts` considered all contexts as custom, including those explicitly specified on the model.


lukeasrodgers:
Actually, I think this fix may be incorrect, I'm going to close this PR until I can verify.

lukeasrodgers:
Never mind, I think it's okay, I just ran into an unrelated bug with `concat` when testing this code.

klacointe:
:+1:

klacointe:
@mbleigh up?

Skulli:
see #786 and #257 
thanks for the fix

EvTi:
@lukeasrodgers thnx for fix!

Erowlin:
Hey, thanks for the fix @lukeasrodgers ! 

Also, it looks similar to #723  and #812. 

Thanks a lot !

klacointe:
@mbleigh up ?

travisofthenorth:
+1

lukeasrodgers:
Merged in master to fix merge conflicts. Failing specs are unrelated to changes here, seem to be due to the fact that bundler 2 requires ruby version >= 2.3.0 (https://bundler.io/blog/2019/01/03/announcing-bundler-2.html)

lukeasrodgers:
bundler issue fixed in https://github.com/mbleigh/acts-as-taggable-on/pull/934, though @mbleigh you may prefer different solution (e.g. just drop testing ruby < 2.3)

ivanovv:
@RudyOnRails can you share your controller code workaround for this bug? Thanks!

RudyOnRails:
> @RudyOnRails can you share your controller code workaround for this bug? Thanks!

Sorry @ivanovv - I can't find it ATM. If I remember where it was, I will post!

apuntovanini:
What's the status of this? It seems still broken, isn't it?
```ruby
@taggable.tag_list = ['one', 'two']
@taggable.skill_list = ['react', 'rails']
@taggable.save
```

only saves one of the two lists

apuntovanini:
@seuros any clues on this? Thanks!

rahulsingh321:
+1

seuros:
This PR is conflicting with the latest code.

lukeasrodgers:
This issue seems to be fixed on `master`.

- merging in upstream `master`, and removing my fixes from `lib/` (since those files no longer exist in the same location), and only leaving the spec change as-is, the specs all pass
- additionally, using an example like @apuntovanini provided above, specs all pass

I will close this issue shortly unless someone can reproduce it on `master`.

rahulsingh321:
@lukeasrodgers I don't think its been fixed on `master`, running the latest version and still getting the same issue.

lukeasrodgers:
@rahulsingh321 do you have a failing spec that reproduces the issue? If so I'd be happy to take a look.

rahulsingh321:
> @rahulsingh321 do you have a failing spec that reproduces the issue? If so I'd be happy to take a look.

Really sorry for the confusion here, was getting issue on:
**Ruby - 2.6.5
Version - 5.0**

After upgrading to 7, all worked. Thanks

PR #729, opened 2016-02-08
Bugfix `TagList#concat` with non-duplicates.

lukeasrodgers:
`uniq!` behaves differently depending on whether duplicates are found or not: if they are, the de-duped list is returned, otherwise `nil` is returned.


seuros:
@lukeasrodgers Thank you

PR #734, opened 2016-02-18
Add rspec tests to tagging_spec 

pcupueran:
I completed the test that were pending in the tagging_spec file. This helps to maintain the code.

### DIFF
--- a/spec/acts_as_taggable_on/tagging_spec.rb
+++ b/spec/acts_as_taggable_on/tagging_spec.rb
@@ -49,12 +49,37 @@ describe ActsAsTaggableOn::Tagging do
     ActsAsTaggableOn.remove_unused_tags = previous_setting
   end
 
+  describe '.owned_by' do
+    before do
+      @tagging_2 = ActsAsTaggableOn::Tagging.new
+      @tagger = ActsAsTaggableOn::Tagger.new
+      @tagger_2 = ActsAsTaggableOn::Tagger.new
+    end
+
+    it "should belong to a specific user" do
+      @tagging.taggable = TaggableModel.create(name: "Black holes")
+      @tagging.tag = ActsAsTaggableOn::Tag.new("Physics")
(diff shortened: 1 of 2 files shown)


spec/acts_as_taggable_on/tagging_spec.rb
```diff
@@ -49,15 +49,74 @@
...
+      @tagging_3.tagger = @tagger_2
+      @tagging_3.context = 'Astronomy'
+      @tagging_3.save
+
+    end
+
+    describe '.owned_by' do
+      it "should belong to a specific user" do
+        expect(@tagging).to be_valid
+        expect(@tagging_2).to be_valid
```
dmgarland: I think this might be tested somewhere else... probably don't need here

spec/acts_as_taggable_on/tagging_spec.rb
```diff
@@ -49,15 +49,74 @@
...
+
+        expect(ActsAsTaggableOn::Tagging.owned_by(@tagger).first).to eq(@tagging)
```
dmgarland: I think it would suffice to check only one permutation here...

spec/acts_as_taggable_on/tagging_spec.rb
```diff
@@ -49,15 +49,74 @@
...
+
+    describe '.by_context' do
+      it "should be found by context" do
+        expect(ActsAsTaggableOn::Tagging.by_context('Science').length).to eq(2);
+      end
+    end
+
+    describe '.by_contexts' do
+      it "should find taggings by contexts" do
+        expect(@tagging_3).to be_valid
```
dmgarland: Ditto... because here we're really testing our before block... once you're happy with it take these out I reckon

dmgarland:
I can't reproduce the travis error.. try touching the readme file with a small change and re-run and see if it's a spurious error...

PR #738, opened 2016-02-23
Tag ordering setting the default ordering

pcupueran:
It allows ordering tags by name or by the taggings_counts. This gives the possibility to order by other criteria than the order of tag creation.


blackst0ne:
Can it be merged, please?

jefflyne:
I would like to see this merged also.

kernelsmith:
I'd also like to see this merged, or something similar. I assumed I'd be able to treat it as a PORO and do this:

app/models/tag.rb

```Ruby
class ActsAsTaggableOn::Tag
  default_scope { order(name: :asc) }
end

```

PR #743, opened 2016-04-07
Remove existing selects from relation

emerson-h:
If there are existing, non-standard selects on a relation, the primary key sub-select will fail because it will include them as well as the primary key, which results in too many columns being present in the results.

This change removes any existing select clauses before applying the selection on the primary key.


emerson-h:
This fixes #744

PR #746, opened 2016-04-12
Always process dirty objects

klacointe:
I encountered a strange behavior with models using the default "tags" context.

Without this fix, dirty fields were not reported in object changes when we update its tag list.


klacointe:
close in favor of #728

PR #749, opened 2016-04-16
Fix rails 5.0.0.beta3 deprecation warning on destroy_all

metalelf0:
Rails 5.0.0.beta3 gives the following deprecation message: `DEPRECATION WARNING: Passing conditions to destroy_all is deprecated and will be removed in Rails 5.1. To achieve the same use where(conditions).destroy_all.`. This PR fixes it as suggested.


seuros:
Thank you.

Issue #662, opened 2015-06-17
How not to destroy taggings when a taggable record is destroyed?


KevinBongart:
Hi,

I'm using the [paranoia gem](https://github.com/radar/paranoia) to "soft delete" records. I'd like to keep taggings in the database even if they point to deleted records, and be able to do something like:

```ruby
Movie.with_deleted.tagged_with(@tag)
```

However, it looks like when a record is destroyed, acts-as-taggable-on destroys taggings [via a `dependent: :destroy` callback](https://github.com/mbleigh/acts-as-taggable-on/blob/c393f86419de10cfb98f21a4f64befb2b4e6d325/lib/acts_as_taggable_on/taggable.rb#L83).

I understand it helps keep the database clean of empty tags, but is there a way to disable this behavior and leave a trace of taggings?

Thanks,

Kevin

infamouse:
Hey, Kevin.  I recently monkey patched this gem to work with paranoia and had to take two steps.  First, add `acts_as_paranoid` to the Tagging class:

```ruby
module ActsAsTaggableOn
  class Tagging < ::ActiveRecord::Base
    acts_as_paranoid
  end
end
```

Second, patch the "save_tags" method to use `really_destroy!` instead of just `destroy`:

```ruby
taggings.not_owned.by_context(context).where(tag_id: old_tags).each { |t| t.really_destroy! }
```

Unfortunately because of the way the module is loaded and applied to the models I had to include the entire class in the initializer (where I've put the monkey patch).  If anyone has a way around this, please let me know!  A latent way to override the save_tags method after initializing, maybe?

```ruby
module ActsAsTaggableOn::Taggable
  module Core
    def self.included(base)
      base.extend ActsAsTaggableOn::Taggable::Core::ClassMethods

      base.class_eval do
        attr_writer :custom_contexts
        after_save :save_tags
      end

      base.initialize_acts_as_taggable_on_core
    end

    module ClassMethods
      def initialize_acts_as_taggable_on_core
        include taggable_mixin
        tag_types.map(&:to_s).each do |tags_type|
          tag_type = tags_type.to_s.singularize
          context_taggings = "#{tag_type}_taggings".to_sym
          context_tags = tags_type.to_sym
          taggings_order = (preserve_tag_order? ? "#{ActsAsTaggableOn::Tagging.table_name}.id" : [])

          class_eval do
            # when preserving tag order, include order option so that for a 'tags' context
[...]

PR #751, opened 2016-04-21
Preserve tags on destroy

dennis-ba:
https://github.com/mbleigh/acts-as-taggable-on/issues/662

I have a case where I need to do a "soft delete" of database records and preserve the tags that are associated. I just need to prevent the destroy callbacks from getting created. 

I've created a new configuration parameter called preserve_tags_on_destroy that is defaulted to false.
Setting this to true removes the destroy callbacks from has_many taggings and context_taggings associations in taggable.rb and core.rb.


SylviaChoong:
Hi @dennis-ba 

Currently initializing `ActsAsTaggableOn.preserve_tags_on_destroy = true` in config/application.rb but error undefined method shows up. Can u fix the conflict and merge this branch for the preserve tags method to be usable? Tq.

ccoeder:
Hi, when this will be merged? it's really nice to have this option.

PR #757, opened 2016-05-29
drop support for Rails < 4

krzysiek1507:


krzysiek1507:
@seuros

krzysiek1507:
@seuros

seuros:
I will work on this today. Thanks for the ping.

krzysiek1507:
Rebased

lib/acts_as_taggable_on/compatibility.rb
```diff
@@ -1,11 +1,7 @@
 module ActsAsTaggableOn::Compatibility
```
seuros: probably we don't need this module now.

krzysiek1507: @seuros it's used [here](https://github.com/mbleigh/acts-as-taggable-on/blob/master/lib/acts_as_taggable_on/tagger.rb#L18-L23), so if we want to drop this module, we have to change API or extract parser for options

krzysiek1507:
Thanks!

seuros:
@krzysiek1507 Thank you for your contribution

PR #761, opened 2016-06-10
Use where before destroy_all to fix Rails 5.0.0.rc deprecations

ianfleeton:
Passing conditions to destroy_all is deprecated and will be removed in
Rails 5.1. To achieve the same use where(conditions).destroy_all.


ianfleeton:
Failed to do appraisal tests and also missed the earlier issue #749 -- closing!

PR #770, opened 2016-07-17
Fix Travis builds for Rails 4.0 and 4.1

krzysiek1507:


seuros:
@krzysiek1507 Thank you , awesome work.
I will push the prerelease

PR #772, opened 2016-07-27
add foreign key support for tagging to tags

BenMorganIO:
Was looking at the migrations and noticed that there was no foreign key added for referential integrity. For Rails 4.2 and 5 users, this a nice addition.

For performance, we could also look at adding `on_delete: :cascade` as well. Thoughts?


seuros:
we still support 4.0.
this migration can be added manually.

maybe we should add a note in the readme.

BenMorganIO:
We could add a migration and leave it commented if a user wants to run it? Or we could add an if statement:

```ruby
unless Rails::VERSION::MAJOR == 4 && Rails::VERSION::MINOR == 0
  add_foreign_key ...
end
```

Issue #773, opened 2016-08-08
Migration installation conflict with spree gem


arpitchauhan:
acts-as-taggable-on version: 4.0.0

I'm trying to upgrade spree from 3.0 to 3.1 but there is a conflict while executing

`bundle exec rake railties:install:migrations`

because both spree gem and acts-as-taggable-on try to add migrations with the same name `AddMissingIndexes`

Manual renaming of one of the two migrations is not that helpful because if one runs `railties:install:migrations` is not considered already copied and thus is copied again.

I suggest renaming your migration to something else, eg `AddMissingIndexes`

I may try to create a PR.

seuros:
go ahead please.

arpitchauhan:
I did create a [PR](https://github.com/mbleigh/acts-as-taggable-on/pull/774)

seuros:
Awesome , thanks

PR #774, opened 2016-08-08
Rename migration to avoid conflicts

arpitchauhan:
The AddMissingIndexes migration conflicts with migrations in other
gem(s), eg spree 3.1. The name has been made more specific to avoid
such conflicts.

Fixes issue #773

### DIFF
--- a/db/migrate/6_add_missing_indexes_on_taggings.rb
+++ b/db/migrate/6_add_missing_indexes_on_taggings.rb
@@ -1,4 +1,4 @@
-class AddMissingIndexes < ActiveRecord::Migration
+class AddMissingIndexesOnTaggings < ActiveRecord::Migration
   def change
     add_index :taggings, :tag_id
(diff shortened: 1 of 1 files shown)


arpitchauhan:
@seuros Hi, is it possible to make a new release 4.0.1 with this PR merged?

krzysiek1507:
@arpitchauhan because file name is changed, `rake railties:install:migrations` will copy this migration, so we should check if the index exists.

arpitchauhan:
@krzysiek1507 indeed, I will make a new PR with that change.

krzysiek1507:
@arpitchauhan could you add it for all migrations with changed names?

arpitchauhan:
@krzysiek1507 @seuros I created #779 

What other migrations have changed names?

krzysiek1507:
@arpitchauhan sorry, I've checked and this is only one. I thought there is more.

PR #779, opened 2016-08-17
Guard against indexes already existing

arpitchauhan:
Since the name of a migration file was recently changed to avoid
conflicts, it will be copied over and run again. Thus, it will now be
checked if the index already exists and will be created only if that's
not the case.

Refer #774


arpitchauhan:
@seuros Is it possible to merge this MR?

PR #782, opened 2016-09-25
Fixes #712 (incompatibility with ActiveRecord::Sanitization#quoted_id)

### LINKED ISSUE #712: Current alpha breaks Rails 5 ActiveRecord::Sanitization#quoted_id
The changes to core.rb at https://github.com/mbleigh/acts-as-taggable-on/commit/d7ac3b0c17f992bed903f20a7ce9e143c0397e0e break the quoted_id method in Rails (which, though slated for deprecation, is still used) by putting in a private quote_value method.

Originally reported at https://github.com/rails/rails/issues/22847 but it's not Rails' problem.

Removing the private from acts_as_taggable_on fixes, but there may be a better fix.

gmcnaughton:
Removes use of the `ActiveRecord::Sanitization#quote_value`, which was
a convenience wrapper for `self.class.connection.quote`.

The convenience method's signature changed in Rails 5.0.0, and it has now been removed completely from Rails master.

Fixes #712

See:
- https://github.com/rails/rails/commit/268ee64e6362467fc6d2e54df703c620af78f5fb - signature change
- https://github.com/rails/rails/commit/12a42f79adb972948c7509a0be2971ca2c37e7c3 - method removed

### DIFF
--- a/lib/acts_as_taggable_on/taggable/core.rb
+++ b/lib/acts_as_taggable_on/taggable/core.rb
@@ -107,14 +107,14 @@ module ActsAsTaggableOn::Taggable
             tags_conditions = tag_list.map { |t| sanitize_sql(["#{ActsAsTaggableOn::Tag.table_name}.name #{ActsAsTaggableOn::Utils.like_operator} ?", t]) }.join(' OR ')
           end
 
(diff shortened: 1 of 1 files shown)


pwim:
My application was also affected by this issue. This PR solves it for me, and none of my application's specs failed, so it looks like it doesn't introduce any regressions. The only thing I could see adding to it is a spec that could prevent this regression from happening again in the future.

seuros:
@gmcnaughton Thank you

gmcnaughton:
Thanks @seuros, happy to help!

PR #785, opened 2016-10-08
Reverse check order before adding custom contexts

stackus:
Reversing this order allows the process to move forward without loading
the `taggings` relationship when a context has already been defined for
the model.

This helps when a model is using a `cached_#{context.to_s.singular}_list`
column and the `taggings` are NOT being eager loaded. It can save a lot
of unexpected N+1 cases.


larsklevan:
I am running into this same issue and I came to the same conclusion about how to fix it. +1 to merge this PR.

PR #800, opened 2017-01-18
Fix typo in README

ashishg-qburst:


seuros:
@ashishg-qburst Thank you

PR #806, opened 2017-02-20
Fix `caching_tag_list_on?` bug (#432)

### LINKED ISSUE #432: undefined method `caching_tag_list_on?'
Since #403 was applied, I'm getting an error rendering a tag_list because the caching_xxx_on? method_on method has not been added. This is rails 4.0.2 working against the acts-as-taggable-on master. It seems that using base.connected? doesn't work reliably on initialization and causes errors later if these methods have not been defined.

Stacktrace is not very interesting but here it is:

```
Showing /Users/aaron/sites/mixxter/app/views/admin/sites/_form.html.erb where line #5 raised:

undefined method `caching_tag_list_on?' for #<Class:0x007fc65e0ea6d8>
Extracted source (around line #5):

    wrapper: :bootstrap3_horizontal, defaults: {column_html: {class: "controls col-sm-9"}, label_html: {class: "col-sm-3"} },
    html: {class: "form-horizontal", role: "form"} do |f| %>
    <%= f.input :name %>
    <%= f.input :tag_list %>
    <%= f.input :description, input_html: {class: 'auto'} %>

    <div class="form-group">

Trace of template inclusion: app/views/admin/sites/edit.html.erb, app/views/admin/sites/show.html.erb

Rails.root: /Users/aaron/sites/mixxter

Application Trace | Framework Trace | Full Trace
app/views/admin/sites/_form.html.erb:5:in `block in _app_views_admin_sites__form_html_erb__4128459791633110155_70244962879300'
app/views/admin/sites/_form.html.erb:1:in `_app_views_admin_sites__form_html_erb__4128459791633110155_70244962879300'
app/views/admin/sites/edit.html.erb:9:in `_app_views_admin_sites_edit_html_erb___2528484288122667039_70244963004780'
app/views/admin/sites/show.html.erb:1:in `_app_views_admin_sites_show_html_erb___3959670009797880860_70244963057000'
app/controllers/application_controller.rb:88:in `set_timezone'
```

ProGM:
Fix #432
## Description:
On Rails 5 (I can't test on other environments right now), if you try to access to a cached tag list before the model as initialized for the first time, it fails with:

```
NoMethodError: undefined method `caching_tag_list_on?' for #<Class:0x007ff6aabcbb98>
```

## How to replicate:
You need a table with an existing record with a `cached_tag_list` column, populated with a non-null value.
Running `ModelName.first.tag_list` -> it throws `NoMethodError`.

Instead, calling `ModelName.columns` before this call, it generates no error.

This is because the `ModelName.columns` injects the `ActsAsTaggableOn::Taggable::Cache::InstanceMethods` in runtime.

Check out here:
https://github.com/mbleigh/acts-as-taggable-on/blob/8e64c3d4a81cfbb8af621228eae36a65c1f94501/lib/acts_as_taggable_on/taggable/cache.rb#L38

If you run it manually or initialize a new model (i.e. `ModelName.new`), it gets injected normally.

---

I added a spec to demonstrate the bug and a possible fix.

Let me know if it can be useful ;)

### DIFF
--- a/lib/acts_as_taggable_on/taggable/core.rb
+++ b/lib/acts_as_taggable_on/taggable/core.rb
@@ -275,7 +275,7 @@ module ActsAsTaggableOn::Taggable
       variable_name = "@#{context.to_s.singularize}_list"
       if instance_variable_get(variable_name)
         instance_variable_get(variable_name)
-      elsif cached_tag_list_on(context) && self.class.caching_tag_list_on?(context)
+      elsif cached_tag_list_on(context) && ensure_included_cache_methods! && self.class.caching_tag_list_on?(context)
         instance_variable_set(variable_name, ActsAsTaggableOn.default_parser.new(cached_tag_list_on(context)).parse)
       else
         instance_variable_set(variable_name, ActsAsTaggableOn::TagList.new(tags_on(context).map(&:name)))
@@ -417,6 +417,10 @@ module ActsAsTaggableOn::Taggable
     end
 
     private
+
+    def ensure_included_cache_methods!
+      self.class.columns
+    end
 
     # Filters the tag lists from the attribute names.
     def attributes_for_update(attribute_names)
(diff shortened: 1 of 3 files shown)


digerata:
I can confirm this fixes my problem in #432. Thanks, Piero!

rbritom:
LGTM @seuros

PR #809, opened 2017-03-02
Due to database collisions, retry when finding or creating a tag

### LINKED ISSUE #693: How to deal with race conditions caused by concurrency
`2015-10-27T15:33:42.642Z 3123 TID-oxfjnp9pw WARN: ActsAsTaggableOn::DuplicateTagError: 'orange' has already been taken`

As part of my project I'm working on I use Sidekiq to run a tagger on 25,000 AR-backed models in postgres. With the default concurrency of 25 workers I instantly run into the above problem every single time.

I assume the responsible code is L74 of https://github.com/mbleigh/acts-as-taggable-on/blob/fd28222629ad59f5a3958424e66e5b9f81a5be18/lib/acts_as_taggable_on/tag.rb

which is `existing_tag || create(name: tag_name)`

By the time existing_tag returns false another worker has created the tag and then create fails.

There are a number of solutions to the race condition and I'm happy to implement. But does anyone have an opinion here? Would a patch be accepted?

Something perhaps like insert/select/where not exists?

fearenales:
This PR closes #693 

Since the [link](https://github.com/mbleigh/acts-as-taggable-on/issues/693#issuecomment-251493235) posted by @adam-e-trepanier is returning 404, I just implemented a retrial strategy described by Mark Quezada in this [blog post](http://blog.mirthlab.com/2012/05/25/cleanly-retrying-blocks-of-code-after-an-exception-in-ruby/).

Any suggestions are welcome! :)

### DIFF
--- a/lib/acts_as_taggable_on/tag.rb
+++ b/lib/acts_as_taggable_on/tag.rb
@@ -69,17 +69,20 @@ module ActsAsTaggableOn
 
       return [] if list.empty?
 
-      existing_tags = named_any(list)
-
       list.map do |tag_name|
-        comparable_tag_name = comparable_name(tag_name)
-        existing_tag = existing_tags.find { |tag| comparable_name(tag.name) == comparable_tag_name }
         begin
+          tries ||= 3
+
+          existing_tags = named_any(list)
+          comparable_tag_name = comparable_name(tag_name)
+          existing_tag = existing_tags.find { |tag| comparable_name(tag.name) == comparable_tag_name }
           existing_tag || create(name: tag_name)
         rescue ActiveRecord::RecordNotUnique
-          # Postgres aborts the current transaction with
-          # PG::InFailedSqlTransaction: ERROR:  current transaction is aborted, commands ignored until end of transaction block
-          # so we have to rollback this transaction
+          if (tries -= 1).positive?
+            ActiveRecord::Base.connection.execute 'ROLLBACK'
+            retry
+          end
+
           raise DuplicateTagError.new("'#{tag_name}' has already been taken")
         end
       end


rbritom:
@fearenales Ill add a test to this

seuros:
@rbritom Please add a reference in the Changelog too .

jdelStrother:
I think this approach causes silent data loss.

If you have two simultaneous processes which call `TaggableModel.create!(tag_list: "hashtag")`, afterwards you'd expect to have 2 TaggableModels stored in the database.
If one of those hits the rollback logic introduced in this PR, you've rolled back the `INSERT INTO taggable_models`, and nothing outside of the retry-loop knows anything about it.  You're left with 1 TaggableModel and no indication that anything went wrong.

@fearenales WDYT?

gmcnaughton:
I share @jdelStrother 's concern about data loss.

The `ROLLBACK` here will roll back whatever transaction you're in, undoing any changes made to the db during that transaction. For example:

```
transaction do
  u = User.create(...)
  u.tag_list.add("foo") # <= a collision here would undo the User.create
end
```

Out of curiosity, what is the reason for the `ROLLBACK`?

jakswa:
Very surprising to me that this `ROLLBACK` is still in this gem since 2018, with multiple concerns/issues raised over the years. I'm thinking now that I gotta hunt for a way to remove this gem or replace it with another 😩.

jakswa:
> @jakswa or open a PR to fix it.

@seuros it is open https://github.com/mbleigh/acts-as-taggable-on/pull/1081

jakswa:
Thank you. Sorry if my original comment came off too negative -- we have been troubleshooting borked transactions that led us to this `ROLLBACK` throwing away data sometimes (but reliably/persistent). If you have two new records getting tagged the same, new tag, you are risking one of those records being `ROLLBACK`'d silently :-\

Issue #811, opened 2017-03-06
Dirty doesn't work when taggings with the same context was previously assigned


eugen0329:
```ruby
begin
  require "bundler/inline"
rescue LoadError => e
  $stderr.puts "Bundler version 1.10 or later is required. Please update your Bundler"
  raise e
end

gemfile(true) do
  source "https://rubygems.org"
  gem "sqlite3"
  gem 'acts-as-taggable-on', '~> 4.0'
  gem 'rails', '4.2.7.1'
end

require "rails"
require 'acts-as-taggable-on'
require "minitest/autorun"

class DummyApp < Rails::Application
  config.root = File.dirname(__FILE__)
  config.autoload_paths << File.dirname(__FILE__)
  config.session_store :cookie_store, key: "cookie_store_key"
  secrets.secret_token    = "secret_token"
  secrets.secret_key_base = "secret_key_base"

  config.logger = Logger.new($stdout)
  Rails.logger  = config.logger
end
Rails.initialize!

ActiveRecord::Base.establish_connection(adapter: "sqlite3", database: ":memory:")
ActiveRecord::Schema.define do
  create_table :users, force: true

  create_table "taggings", force: :cascade do |t|
    t.integer  "tag_id"
    t.integer  "taggable_id"
    t.string   "taggable_type"
    t.integer  "tagger_id"
    t.string   "tagger_type"
    t.string   "context",       limit: 128
    t.datetime "created_at"
  end

  create_table "tags", force: :cascade do |t|
    t.string  "name"
    t.integer "taggings_count", default: 0
  end
end
class User < ActiveRecord::Base
  acts_as_taggable_on :segments
end

class BugTest < Minitest::Test
  def setup
    User.new.tap { |u| u.segment_list = 'a' }.save

    @user = User.last
  end

  def test_changes_not_blank
    @user.segment_list = ['a', 'b']
    assert @user.changes.present?
    assert @user.segment_list_changed?
  end
end

```

I suppose, the root of the issue is `unless custom_contexts.include?(context.to_s)` condition at [line](https://github.com/mbleigh/acts-as-taggable-on/blob/master/lib/acts_as_taggable_on/taggable/core.rb#L327).

deep-spaced:
I'm getting the same sort of error in my Rails app: changing a `tag_list` on an existing record doesn't trigger a dirty status. @eugen0329, your example above illustrates this well. 

[...]

PR #812, opened 2017-03-06
Fix dirty triggering with existing custom context

eugen0329:
This PR solves #811 issue.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ As such, a _Feature_ would map to either major or minor. A _bug fix_ to a patch.
  * Added
    * [@jessieay  Rails 5 ](https://github.com/mbleigh/acts-as-taggable-on/pull/763)
  * Fixes
+  * [@eugen0329 Fix dirty triggering with existing custom context] (https://github.com/mbleigh/acts-as-taggable-on/issues/811)
   * [@rikettsie #623 collation parameter is ignored if it generates an exception] (https://github.com/mbleigh/acts-as-taggable-on/pull/650 )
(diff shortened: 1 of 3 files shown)


eugen0329:
Oh, everything was well locally :(

BTW the problem with dirty status still remains. Can you, please, clarify what was the original purpose of a custom context check [here](https://github.com/mbleigh/acts-as-taggable-on/pull/812/files#diff-6fb8d10ba793f1b09e64f2923988ee4eL327)?

Erowlin:
Hey, thanks for the fix @eugen0329 !

Also, it looks similar to #723 and #728.

Thanks a lot !

marlonmarcos21:
@mbleigh Are you looking into this? This is still broken and this PR actually fixes it. Specs are fine locally for me too.

travisofthenorth:
+1. This is a pretty big issue and it would be great to have it fixed.

seuros:
This PR is not current anymore.

ivanovv:
@seuros I'm sorry to ask in a closed issue, but is there a version of this gem that works fine with Rails 4.2? We are quite lagging with updates to Rails 5, hence my question.

Also, is there a workaround for #811 that I can implement in app code?

seuros:
@ivanovv I think version 5.0.0 has the fix. Did you try it ?

ivanovv:
@seuros my bad, I've read this commit message

> Drop support for old version of ActiveRecord and Ruby

and made an incorrect assumption (Rails v5 or later)

Will give it a go!

ivanovv:
Looks like the fix never made it into v5 https://github.com/mbleigh/acts-as-taggable-on/compare/v4.0.0...v5.0.0

My specs also fail.

Issue #700, opened 2015-11-09
Why is context on taggable table?


NeoElit:
Why is context is entered each time in taggings table instead of putting it in the tags table?
Instead of repeating context on each tagging entry, would it be better if it is added and separate tags are created for different contexts?

seuros:
because a tag can belong to many contexts.

NeoElit:
yeah that right. But my query is, isn't it good to create tag table with context (different tag_id for unique [tag_name, tag_context]) instead of repeating string context in each tagging ? Normally tagging entries will be out numbered by tag entries.

walidvb:
I agree with @NeoElit NeoElit, and have a usecase where i would rather have contexts in the `tags` table. I don't really see where i would want to share the same tag across contexts. 

I have a  `Book` and a `User` model. `Book`s are taggable_on `genres`, and `User`s on `fav_genres`.
By merging the two contexts, it is difficult to get only the genres of all `Book`s, or the count of all _history_  books(the count would also return the number of users who have _history_ as a `fav_genre`).

What is the usecase of the current implementation? If i have a user that has `skills = 'french'` **and** `languages='french'`, i would not want the tag `french` to have 2 counts, imo.

digerata:
One of the things you can't do without context on the tags table is present a list of tags to users based on the context. Being able to do this is a huge usability win. E.g., "Here are tags you can use for hobbies and it won't show the tags used for movie genres."

(In fact, we maintain a fork of this library with the tags table having a context column.)

razum2um:
@digerata please take a look on https://github.com/mbleigh/acts-as-taggable-on/pull/818 - will it solve issue for you?

PR #818, opened 2017-03-22
Ability to use tags independently from different models

razum2um:
This PR is based on 2 issues:

1) when used on different models, tags get unified by name and `taggings_count` shows sum across all models 
2) for a simple thing like "get all tags applicable to 1 model, not to others" we have to aggregate through taggings like group by type, use ids to select tag records and counts instead of taggings_count.

After PR we can use an option to split tags exclusively for a model
```ruby
class Order < ActiveRecord::Base
  acts_as_taggable exclusive: true # this will isolate order tags completely
end

# and use additional dsl to get needed tags only
ActsAsTaggableOn::Tag.exclusively_for(:order)
```

For further details please, see updated readme and a new [spec](https://github.com/razum2um/acts-as-taggable-on/blob/master/spec/acts_as_taggable_on/acts_as_taggable_on_spec.rb#L304L324) :)

P.S. I saw [find_or_create_tags_from_list_with_context](https://github.com/mbleigh/acts-as-taggable-on/blob/master/lib/acts_as_taggable_on/taggable/core.rb#L433L454) in the code, but I think this hook looks kinda ugly, exclusively option makes this useless, as you can pass a `tags_class` directly and it will work out of the box.

Do you like the idea of PR?
I also thought about making `type` column optional and commentabe/uncommentable in migration for users which don't wont to migrate possible huge tables, but WDYT about default behaviour?

### DIFF
--- a/README.md
+++ b/README.md
@@ -424,6 +424,49 @@ CSS:
 .css4 { font-size: 1.6em; }
 ```
 
+### Exclusive tags depending on models
+
+Let's say you need a complete independent tag lists for 2 of your domain models,
+say `Order` and `Product`, so that they couldn't clash. Use `exclusive` parameter for that
+
+```ruby
+class Order < ActiveRecord::Base
+  acts_as_taggable exclusive: true
+end
+
+class Product < ActiveRecord::Base
+  acts_as_taggable exclusive: true
+end
+```
+
+And operate on orders and tags complete like usual, except tags will not clash:
+
+```ruby
+order = Order.create!
+order.tag_list = 'cool, nice, wow'
+order.save!
+
+product = Product.create!
+product.tag_list = 'super, nice'
+product.save!
+
+ActsAsTaggableOn::Tag.where(name: 'nice').count
+# => 2
+
+ActsAsTaggableOn::Tag.where(name: 'nice').pluck(:type)
+# => ["ActsAsTaggableOn::Tag::OrderTag", "ActsAsTaggableOn::Tag::ProductTag"]
+```
+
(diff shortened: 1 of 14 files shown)


syguer:
Have you tried to use `context` ?
https://github.com/mbleigh/acts-as-taggable-on/blob/master/lib/acts_as_taggable_on/taggable.rb#L39-L41

I think almost use case is cleared by using `context` .
If you need `excluding` after using `context`,  I think you should implement the method for negative condition of below.
https://github.com/mbleigh/acts-as-taggable-on/blob/master/lib/acts_as_taggable_on/tag.rb#L51-L55

Thanks.

razum2um:
@syguer you're partially right, but the drawback of `context` is that it's located on `taggings`, but really we need it (or I suggest `type` on `tags`)

Please see https://github.com/mbleigh/acts-as-taggable-on/issues/700 people already complaining about that and forking because of this

syguer:
I've understood it is convenient if tags have context( or type, kind).
Idea looks good. But my concern is It looks complicated and decrease maintainability.
Maybe, best solution is it is able to create some tags-taggings set.
I want to hear maintainer's opinion.

razum2um:
@seuros but why? it's not stale, just 9yrs passed :)

seuros:
_**This PR is clearly AI generated.**_  🧌 


Seriously :), the API changed a lot , i can't merge it without breaking the current contract.

This you try NoFlyList gem ? It a drop-in replacement .

(* I'm grooming the backlog , to merge the fixes only)

PR #821, opened 2017-04-20
Fix a bug on ActAsTaggableOn::Tag.find_or_create_with_like_by_name

syguer:
Hi :)
I found a weird issue.
Passing too long name(over 255 characters) to `ActAsTaggableOn::Tag.find_or_create_with_like_by_name`, it returns not persisted instance.
Because inside `.create` method returns not persisted instance when instance is invalid.

This issue occurs some problems.
For example,  `ActsAsTaggableOn::Taggable#save_tags` fails with 'Tag can't be blank'. It is hard to find [reason](https://github.com/mbleigh/acts-as-taggable-on/blob/v4.0.0/lib/acts_as_taggable_on/taggable/core.rb#L420).

This patch fixes that issues.
Please check this out and release to rubygems as soon as possible.
Thanks 👻

### DIFF
--- a/lib/acts_as_taggable_on/tag.rb
+++ b/lib/acts_as_taggable_on/tag.rb
@@ -60,7 +60,7 @@ module ActsAsTaggableOn
       if ActsAsTaggableOn.strict_case_match
         self.find_or_create_all_with_like_by_name([name]).first
       else
-        named_like(name).first || create(name: name)
+        named_like(name).first || create!(name: name)
       end
     end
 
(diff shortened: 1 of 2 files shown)


lib/acts_as_taggable_on/tag.rb
```diff
@@ -60,7 +60,7 @@ def self.find_or_create_with_like_by_name(name)
       if ActsAsTaggableOn.strict_case_match
         self.find_or_create_all_with_like_by_name([name]).first
       else
-        named_like(name).first || create(name: name)
+        named_like(name).first || create!(name: name)
```
seuros: this will raise a exception .

syguer: hmm.. OK.
I'll change approach.

PR #830, opened 2017-05-18
`cached_tag_list` column is ignored

ProGM:
Just a spec demonstrating that the `cached_tag_list` column is ignored when calling `tag_list`.

The strange thing is that, instead, `tag_list_cache_on(:tags)` works as expected.

Still investigating.

The failing line is this one:
https://github.com/mbleigh/acts-as-taggable-on/pull/830/files#diff-0bfae9e63c455d336360aad3c5266d9aR27

### DIFF
--- a/lib/acts_as_taggable_on/taggable/core.rb
+++ b/lib/acts_as_taggable_on/taggable/core.rb
@@ -107,8 +107,12 @@ module ActsAsTaggableOn::Taggable
       self.class.grouped_column_names_for(object)
     end
 
+    def unpersisted_custom_contexts
+      @unsaved_custom_contexts ||= []
+    end
+
     def custom_contexts
-      @custom_contexts ||= taggings.map(&:context).uniq
+      @custom_contexts ||= (taggings.map(&:context) + unpersisted_custom_contexts).uniq
     end
 
     def is_taggable?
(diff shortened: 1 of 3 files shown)


lib/acts_as_taggable_on/taggable/core.rb
```diff
@@ -107,8 +107,12 @@ def grouped_column_names_for(object)
       self.class.grouped_column_names_for(object)
     end
 
+    def unpersisted_custom_contexts
+      @unsaved_custom_contexts ||= []
```
rbritom: @ProGM sorry, why is the method name and the instance variable name different?

lib/acts_as_taggable_on/taggable/core.rb
```diff
@@ -107,8 +107,12 @@ def grouped_column_names_for(object)
...
+    end
+
     def custom_contexts
-      @custom_contexts ||= taggings.map(&:context).uniq
+      @custom_contexts ||= (taggings.map(&:context) + unpersisted_custom_contexts).uniq
```
rbritom: I think you have an opportunity here to do ```taggings.pluck(:context).uniq``` and it will be a query, I think it would be good to delegate to the RDBMS as much as possible. What do you think? @ProGM

lib/acts_as_taggable_on/taggable/core.rb
ProGM: oops, typo! Fixing...

lib/acts_as_taggable_on/taggable/core.rb
ProGM: Uhm, if you use `pluck`, then the taggings model won't be cached. So, accessing to it again will result in another query.

Not sure if it's better or worse :(

rbritom: @ProGM hmm, Ill have a closer look then.

Issue #838, opened 2017-06-21
Performance issues when adding or updating tags


bduran82:
Running on Rails 4.2.8 and gem version 5.0.0 we experience a poor performance when updating the tags of a model with big number of tags.

In our project we need to handle models with sometimes big number of tags (up to around 300). We have experienced performance issues in our database when updating tags in those models that even led to request timeouts. The source seems to be the huge amount of database queries like the following:

```SELECT `tags`.* FROM `tags` WHERE (name = BINARY 'editor' OR name = BINARY 'funding' OR name = BINARY 'transfers' OR name = BINARY 'role' OR name = BINARY 'editorial' OR ...```

It looks like in `ActsAsTaggableOn::Tag find_or_create_all_with_like_by_name` all the tags attached to the model are repeatedly reloaded for every tag to avoid collisions in the database. Shouldn't the tags be loaded once in the beginning of the method and then only refreshed again in case there is a collision?

```
  def self.find_or_create_all_with_like_by_name(*list)
      list = Array(list).flatten

      return [] if list.empty?

      list.map do |tag_name|
        begin
          tries ||= 3

          existing_tags = named_any(list)
          comparable_tag_name = comparable_name(tag_name)
          existing_tag = existing_tags.find { |tag| comparable_name(tag.name) == comparable_tag_name }
          existing_tag || create(name: tag_name)
        rescue ActiveRecord::RecordNotUnique
          if (tries -= 1).positive?
            ActiveRecord::Base.connection.execute 'ROLLBACK'
            retry
          end

          raise DuplicateTagError.new("'#{tag_name}' has already been taken")
        end
      end
    end
```

seuros:
This was fixed in a later release. If it still happens on v13 with Rails 8, please open a new issue with a repro.

PR #839, opened 2017-06-22
Avoid unnecessary queries when finding or creating tags

bduran82:
This is an optimisation to avoid database queries when finding or creating a list tags in `ActsAsTaggableOn::Tag find_or_create_all_with_like_by_name`:

 The list of tags to be retrieved is only updated once and refreshed again in case there is a tag name collision.

This solves [#838](https://github.com/mbleigh/acts-as-taggable-on/issues/838) issue

### DIFF
--- a/lib/acts_as_taggable_on/tag.rb
+++ b/lib/acts_as_taggable_on/tag.rb
@@ -69,17 +69,17 @@ module ActsAsTaggableOn
 
       return [] if list.empty?
 
+      existing_tags = named_any(list)
       list.map do |tag_name|
         begin
           tries ||= 3
-
-          existing_tags = named_any(list)
           comparable_tag_name = comparable_name(tag_name)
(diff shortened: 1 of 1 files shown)


Erowlin:
Very much needed actually.

jnimety:
I'm running into this performance problem with large numbers of tags too, the patch seems to fix the issue.

gen0cide:
Any idea when this could be merged into v5?

toobulkeh:
Would love a merge on this

bitops:
👍  on this getting merged.

Erowlin:
Amazing, thank you @bitops

Cremz:
This really needs to happen.
It's not only when there are many tags, even 5-10 is enough to make everything seems slow.
It makes one query for each tag, each query takes about 2 seconds. ( i have about 130k tags already). it's not efficient at all. I try to add one record, that takes 5 ms, but updating the tags takes up to 10 seconds

PR #840, opened 2017-06-26
Upgrade to Rails 5.1.1

pavel-jurasek-hypofriend-de:


seuros:
uuid is not supported by mysql and sqlite3. And ApplicationRecord should not be used in gems.

Thank you for your countribution

PR #843, opened 2017-06-28
Lower required Ruby version to 2.2.2

tvdeyen:
Rails 5 itself has this as lowest Ruby version allowed.

If we keep the 2.2.7 as before we break a lot of Rails apps
and Rails engines that allow Ruby 2.2.2.

I didn't find a reason to not support 2.2.2.


dwarner:
@mbleigh Was there a reason for the 2.2.7 requirement?

PR #847, opened 2017-07-18
Bugfix: 'context' does not work properly under a specific condition

pawa-:
If `taggable` has the following tags:
```
taggable = TaggableModel.new(name: 'Bob Jones')
taggable.skill_list = 'ruby, rails, css, julia'
taggable.tag_list = 'bob, charlie, julia'
taggable.save
```

`TaggableModel.tagged_with('julia', on: :tags).size` returns 2 instead of 1.

The bug has been introduced by https://github.com/mbleigh/acts-as-taggable-on/commit/64ed839a8a6021c719fbff9c6bdf74a4fcd65f18


dnahodil:
Thanks for putting this fix together @pawa- 👍 

Hi @seuros, it would be great to get this bug fix accepted if possible. Thanks!

dnahodil:
Thanks @seuros

PR #849, opened 2017-08-02
Revert "Added missed indexes."

SuperTux88:
This is the same as in #709, but the file [was not removed when merging this](https://github.com/mbleigh/acts-as-taggable-on/commit/5163a75090f82916613c60d940477c678952a065), because it [was renamed in the meantime](https://github.com/mbleigh/acts-as-taggable-on/commit/8aa9f8f17c05424d061b92b3b8acd9aa5537064b). So remove it again.


SuperTux88:
@seuros why did you close this? the file still exists.

PR #850, opened 2017-08-04
Update README.md installation guide (5.0 released)

JarLowrey:


JarLowrey:
This package hasn't been touched for 7 months, I'd probably it dead and would be wary of using it.

PR #859, opened 2017-09-15
Allow multiple contexts to be specified in for_context

travisofthenorth:
This PR contains a minor change to `for_context` that allow multiple values to be provided. When multiple values are provided, it uses an `IN` clause.


travisofthenorth:
@seuros any chance you could take a look at this?

Issue #836, opened 2017-06-14
[Rails 5.1] Fix dirty.rb to use the new API.  *_changed?, *_was...


Erowlin:
This https://github.com/mbleigh/acts-as-taggable-on/blob/8e64c3d4a81cfbb8af621228eae36a65c1f94501/lib/acts_as_taggable_on/taggable/dirty.rb should be rewritten to use the new Rails 5.1 API Dirty. 

Here is the list of changes introduced in Rails 5.1: 
https://github.com/rails/rails/pull/25337#issuecomment-225166796

Relevant ones for acts-as-taggable:
- `def #{tag_type}_list_changed?` -> `def will_save_change_to_#{tag_type}_list?`
- `def #{tag_type}_list_was` -> `def #{tag_type}_list_in_database_was`
- `def #{tag_type}_list_change` -> `def #{tag_type}_list_ change_to_be_saved`
- `def #{tag_type}_list_changes` -> **Not sure about that one.** 

Otherwise, it's gonna trigger **a lot** of deprecation warnings : 
```
DEPRECATION WARNING: The behavior of `attribute_changed?` inside of after callbacks will be changing in the next version of Rails. The new return value will reflect the behavior of calling the method after `save` returned (e.g. the opposite of what it returns now). To maintain the current behavior, use `saved_change_to_attribute?` instead.
```

Erowlin:
Actually it's broken. 
```ruby
user.my_tag_list
#=> ['hello'] 
user.my_tag_list = 'world'
user.my_tag_list_changed?
#=> false
```
While it should returns true.

haraldkrischner:
I can confirm this. Have you thought about a workaround?

Erowlin:
Yes we found a dirty hack which is basically in our controller

```ruby
previous_tag_list = user.my_tag_list.to_set
user.assign_attributes(params)
do_something if previous_tag_list != user.my_tag_list.to_set
```

Without this hack, it would look like that: 
```ruby
user.assign_attributes(params)
do_something if user.will_save_change_to_my_tag_list
```

haraldkrischner:
Thanks for your quick answer, will have a look into it

Erowlin:
Another very very important note about that, which I did not mentioned in my original comment: actually the original model is not dirtified also, while it should be: 
```ruby

user. my_tag_list = ['123']
# Using the old API for the example, note that `changed?` has been replaced by `has_changes_to_save`
user.my_tag_list.changed?
[...]

PR #861, opened 2017-09-20
attribute_changed? ins deprecated as of Rails 5.1

klacointe:
`attribute_changed?` is now deprecated and replaced by `will_save_change_to_attribute?`

Also, caling `changed_attributes` method generates a deprecation warning.

### DIFF
--- a/lib/acts_as_taggable_on/taggable/dirty.rb
+++ b/lib/acts_as_taggable_on/taggable/dirty.rb
@@ -13,19 +13,20 @@ module ActsAsTaggableOn::Taggable
 
           class_eval <<-RUBY, __FILE__, __LINE__ + 1
             def #{tag_type}_list_changed?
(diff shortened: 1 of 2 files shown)


Erowlin:
Duplicate #836. I suggest you bump the ticket.

klacointe:
@mbleigh up?

lunaru:
@klacointe since `will_save_change_to_tag_list?` is not defined in Rails versions below 5.1 this PR would also require bumping the dependencies of this library to be targeted to Rails 5.1 at a minimum.

klacointe:
@lunaru since `will_save_change_to_tag_list?` is just an alias for ` :tag_list_changed?` this should works in Rails below 5.1

seuros:
We dont use this file anymore.
Thanks for your contribution

PR #862, opened 2017-09-21
Rails 5.1 changed_attributes behaviour when working on assocation fixed

haraldkrischner:
fixing #836


haraldkrischner:
I could use help with the rails 4.2 part :-\

Issue #869, opened 2017-11-11
tagged_with searches and inconsistent case sensitivity


tekniklr:
- acts-as-taggable-on (5.0.0)
- rails (5.1.4)
- activerecord-oracle_enhanced-adapter (1.8.2)
- ruby 2.4.2

When searching using `Model.tagged_with('Tagname')`, nothing is returned, despite these taggings existing. Comparing the SQL generated with `Model.tagged_with('Tagname', any: true)` revealed why- the latter forces lower case in SQL, the former does not. Since the actual tag name is capitalized in the database, it can never be returned in search without searching for `any`.

**without `any: true` (incorrectly case sensitive):**
`SELECT "NOTES".* FROM "NOTES" INNER JOIN "TAGGINGS" "NOTE_TAGGINGS_8D57F90" ON "NOTE_TAGGINGS_8D57F90"."TAGGABLE_ID" = "NOTES"."ID" AND "NOTE_TAGGINGS_8D57F90"."TAGGABLE_TYPE" = 'Note' AND "NOTE_TAGGINGS_8D57F90"."TAG_ID" IN (SELECT "TAGS"."ID" FROM "TAGS" WHERE "TAGS"."NAME" LIKE 'concerns' ESCAPE '!')`

**with `any: true` (no case sensitivity- everything is lowercased in SQL):**
`SELECT "NOTES".* FROM "NOTES" WHERE (EXISTS (SELECT * FROM "TAGGINGS" WHERE "TAGGINGS"."TAGGABLE_ID" = "NOTES"."ID" AND "TAGGINGS"."TAGGABLE_TYPE" = 'Note' AND "TAGGINGS"."TAG_ID" IN (SELECT "TAGS"."ID" FROM "TAGS" WHERE (LOWER("TAGS"."NAME") LIKE 'concerns' ESCAPE '!'))))`

Running the first query in a SQL console and changing `SELECT "TAGS"."ID" FROM "TAGS" WHERE "TAGS"."NAME" LIKE 'concerns' ESCAPE '!'` to `SELECT "TAGS"."ID" FROM "TAGS" WHERE (LOWER("TAGS"."NAME") LIKE 'concerns' ESCAPE '!'` works as expected.

PR #871, opened 2017-11-15
matches_attribute was not being used in tag_match_type

tekniklr:
addresses https://github.com/mbleigh/acts-as-taggable-on/issues/869


botandrose:
I ran into #869 on upgrading to 5.0.0, and can confirm this patch fixed it for me, as well. Suggest merging,  and releasing a 5.0.1?

PR #873, opened 2017-12-07
Fix tagged_with matching wrong contexts

jdelStrother:
I think tagged_with has a context-matching bug since the rewrite.  Given three tag-contexts, "apple", "banana", and "candy", `tagged_with("foo", on: "banana")` can return taggables that are in tag-contexts of "apple" or "banana".

It's matching all taggings where `context <= "banana"`, rather than where `context = "banana"`.
I'm guessing this is a copy-paste typo (/cc @rbritom, @seuros ?)


jdelStrother:
Oh.   See #847 😕

jdelStrother:
Fixed by @pawa-

PR #876, opened 2018-01-19
Avoid unnecessary queries for custom_tags

fatkodima:


lib/acts_as_taggable_on/taggable/core.rb
```diff
@@ -108,15 +108,15 @@ def grouped_column_names_for(object)
     end
 
     def custom_contexts
-      @custom_contexts ||= taggings.map(&:context).uniq
+      @custom_contexts ||= taggings.pluck("DISTINCT context")
```
fatkodima: No need to load all taggings into memory for this

lib/acts_as_taggable_on/taggable/core.rb
```diff
@@ -108,15 +108,15 @@ def grouped_column_names_for(object)
...
     end
 
     def is_taggable?
       self.class.is_taggable?
     end
 
     def add_custom_context(value)
-      custom_contexts << value.to_s unless custom_contexts.include?(value.to_s) or self.class.tag_types.map(&:to_s).include?(value.to_s)
+      custom_contexts << value.to_s unless self.class.tag_types.map(&:to_s).include?(value.to_s) || custom_contexts.include?(value.to_s)
```
fatkodima: Just checks in-memory `tag_types` first and then, if not found, query the db

PR #879, opened 2018-01-28
fix dirty behaviour for items with persisted taggings

### LINKED ISSUE #811: Dirty doesn't work when taggings with the same context was previously assigned

iiwo:
Fixes #811. 

If item contains persisted taggings dirty methods will not work for the newly instantiated object:
```
item = Item.create(tag_list: 'a')
item = Item.find(item.id)
item.tag_list = 'b, c'
item.tag_list_changed? # => false
```

Root cause of this is the `custom_contexts` method returning all taggings contexts instead of only the custom ones, in result the `process_dirty_object` method is not called if any persisted tagging exist for the item for that context

this is similar to #723 and #728 but based on the current codebase and only addressing this specific issue with minimal impact to other/unrelated functionalities. The `custom_contexts` method is only used as a guard for process_dirty_object and preventing duplicates in `add_custom_context` and this behaviour is not affected, so this should be safe to merge

**NOTE:** the failures in Travis build are not related - it should all pass after #880 is merged

### DIFF
--- a/lib/acts_as_taggable_on/taggable/core.rb
+++ b/lib/acts_as_taggable_on/taggable/core.rb
@@ -108,7 +108,7 @@ module ActsAsTaggableOn::Taggable
     end
 
     def custom_contexts
-      @custom_contexts ||= taggings.map(&:context).uniq
+      @custom_contexts ||= taggings.map(&:context).uniq - declared_contexts
     end
 
     def is_taggable?
@@ -116,7 +116,7 @@ module ActsAsTaggableOn::Taggable
     end
 
     def add_custom_context(value)
-      custom_contexts << value.to_s unless custom_contexts.include?(value.to_s) or self.class.tag_types.map(&:to_s).include?(value.to_s)
+      custom_contexts << value.to_s unless custom_contexts.include?(value.to_s) or declared_contexts.include?(value.to_s)
     end
 
     def cached_tag_list_on(context)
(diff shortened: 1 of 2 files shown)


jsquirrelz:
+1

Definitely agree that the expected behavior of `item.tag_list_changed?` should return `true` when new items are added to the `tag_list` and has not been saved. Confirmed this branch fixes the current behavior.

travisofthenorth:
+1 would love to see this fixed

iiwo:
seems to be fixed in latest versions and no longer relevant after dropping support for AR <5

bolshakov:
No support, no problem

PR #880, opened 2018-01-28
add version requirement for pg gem to address rails/rails#31671

iiwo:
add version requirement for pg gem - fixes postgres builds currently failing because of https://github.com/rails/rails/pull/31671


seuros:
We are dropping support for AR 4.2 in the next version.

Thank you for your contribution

Issue #864, opened 2017-10-04
Tags with underscores get escaped


chevinbrown:
Being a ruby shop, we'd like to have tags that use underscores instead of dashes. ie `big_pandas` instead of `big-pandas`.

Looking at the utils, for postgres, the escape character is `!`, but shouldn't it be `\\`?

Will be happy to submit a PR to fix this if you agree.

rsl:
would love to see this as well

lunaru:
@chevinbrown are you guys still planning to put up a PR for this?

chevinbrown:
@rsl @lunaru we haven't started one, but looks like it would be appreciated by the community. @dwilkins this would be a good team effort.

rsl:
Welcome comments on my PR, everyone. Thanks.

PR #882, opened 2018-02-08
improve postgres escape sequence

rsl:
Fixes https://github.com/mbleigh/acts-as-taggable-on/issues/864

I'm not sure why ActsAsTaggableOn::Utils#escape_like is named specifically for like when it seems to be generic escape functionality. wondering if this should be renamed to just escape?

Also, unsure about add myself to CHANGELOG.md and documenting this change since all the entries seem to be tied to a gem version. Is this no longer the practice? Could not see this in other recent PR as well.

Thanks and sorry for all the questions heh.

### DIFF
--- a/lib/acts_as_taggable_on/taggable/tagged_with_query/query_base.rb
+++ b/lib/acts_as_taggable_on/taggable/tagged_with_query/query_base.rb
@@ -48,7 +48,7 @@ module ActsAsTaggableOn::Taggable::TaggedWithQuery
 
     def escaped_tag(tag)
       tag = tag.downcase unless ActsAsTaggableOn.strict_case_match
-      tag.gsub(/[!%_]/) { |x| '!' + x }
+      ActsAsTaggableOn::Utils.escape_like(tag)
     end
 
     def adjust_taggings_alias(taggings_alias)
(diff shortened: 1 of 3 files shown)


lib/acts_as_taggable_on/taggable/tagged_with_query/query_base.rb
```diff
@@ -48,7 +48,7 @@ def tags_match_type
 
     def escaped_tag(tag)
       tag = tag.downcase unless ActsAsTaggableOn.strict_case_match
-      tag.gsub(/[!%_]/) { |x| '!' + x }
+      ActsAsTaggableOn::Utils.escape_like(tag)
```
rsl: The previous code was identical to ActsAsTaggableOn::Utils.escape_like and it seemed DRYer to funnel this all through a singular interface. But this makes the name seem odd now too.

lunaru:
👍 Looks great! Would love to see this get merged in. Build is failing for the same reason as already addressed in this PR https://github.com/mbleigh/acts-as-taggable-on/pull/880

rsl:
thanks

rsl:
@seuros hi. this got rolled back because it was breaking build. it passes for me locally. is there a way to run it through the same CI as the main repo?

seuros:
Of course, if you open a PR, Travis will run on it.

rsl:
managed to find the travis build for this PR! https://travis-ci.org/mbleigh/acts-as-taggable-on/jobs/339031800 

it looks like there's a configuration issue? https://travis-ci.org/mbleigh/acts-as-taggable-on/jobs/339031800#L657

PR #872, opened 2017-11-30
rails 5.2 compability

doits:
Since rails 5.2, `@changed_attributes` is not available anymore (besides being a frozen hash when calling `changed_attributes`). This lead to:

```
     NoMethodError:
       undefined method `[]=' for nil:NilClass
     # /cache/gems/ruby_2.4/gems/acts-as-taggable-on-5.0.0/lib/acts_as_taggable_on/taggable/core.rb:206:in `process_dirty_object'
     # /cache/gems/ruby_2.4/gems/acts-as-taggable-on-5.0.0/lib/acts_as_taggable_on/taggable/core.rb:184:in `set_tag_list_on'
     # /cache/gems/ruby_2.4/gems/acts-as-taggable-on-5.0.0/lib/acts_as_taggable_on/taggable/core.rb:45:in `tag_list='
```

I quickly worked around the issue by using `attributes_changed_by_setter` which fixes the error and makes things looks like they work (like setting and getting tags etc). Though two specs regarding changed attributes still fail:

```
  1) ActsAsTaggableOn::Taggable::Dirty with un-contexted tags when tag_list changed should show changes of dirty object
     Failure/Error: expect(@taggable.changes).to eq({'tag_list' => ['awesome, epic', ['one']]})

       expected: {"tag_list"=>["awesome, epic", ["one"]]}
            got: {"tag_list"=>["awesome, epic", nil]}

       (compared using ==)

       Diff:
       @@ -1,2 +1,2 @@
       -"tag_list" => ["awesome, epic", ["one"]],
       +"tag_list" => ["awesome, epic", nil]

 2) ActsAsTaggableOn::Taggable::Dirty with context tags when language_list changed should show changes of dirty object
     Failure/Error: expect(@taggable.changes).to eq({'language_list' => ['awesome, epic', ['one']]})

       expected: {"language_list"=>["awesome, epic", ["one"]]}
            got: {"language_list"=>["awesome, epic", nil]}

       (compared using ==)

       Diff:
       @@ -1,2 +1,2 @@
       -"language_list" => ["awesome, epic", ["one"]],
       +"language_list" => ["awesome, epic", nil]
```

This seems to be the same spec for contexed and uncontexec tags, so the same failure. I don't exactly understand why the expected output is `["awesome, epic", ["one"]]` and not `["awesome, epic", "one"]` (no array), but nevertheless `["awesome, epic", nil]` is wrong.

So my fix is not 100 % correct yet, but I did not have time to look into it at more detail.

PR #887, opened 2018-02-24
Rails 5.2 support, drop Rails 4.2 support.

Fodoj:
Based on https://github.com/mbleigh/acts-as-taggable-on/pull/872 with extra changes on top

I've also dropped Rails 4.2 support, as it requires lots of hacking around in the code - `dirty` implementation changed a lot since then.

This PR rolls back https://github.com/mbleigh/acts-as-taggable-on/commit/cc7a2c097ee9eeb4417328986f0dd49112cb05b3 as it breaks all PostgreSQL builds (including master branch).

Upd 26.05.18: ready to merge!


spec/acts_as_taggable_on/dirty_spec.rb
```diff
@@ -123,5 +119,19 @@
         expect(@taggable.changes).to be_empty
       end
     end
+
+    context 'when language_list changed by association' do
+      let(:tag) { ActsAsTaggableOn::Tag.new(name: 'one') }
+
+      before(:each) do
```
Erowlin: I would refactor this to 
```ruby
    it 'flags language_list as changed' do
        expect(@taggable.changes).to be_empty
        @taggable.languages << tag
        expect(@taggable.language_list_changed?).to be_truthy
end
```

Unless there is a specific guideline around it, I usually tend to not over-use the before-each, harder to read. Since here you have only one test in your context, I would drop it.

spec/acts_as_taggable_on/dirty_spec.rb
```diff
@@ -123,5 +119,19 @@
...
+        expect(@taggable.changes).to be_empty
+        @taggable.languages << tag
+      end
+
+      it 'flags language_list as changed' do
+        expect(@taggable.language_list_changed?).to be_truthy
+      end
+    end
+
```
Erowlin: Un-needed new line

lib/acts_as_taggable_on/taggable/core.rb
```diff
@@ -28,12 +28,16 @@ def initialize_acts_as_taggable_on_core
             has_many context_taggings, -> { includes(:tag).order(taggings_order).where(context: tags_type) },
                      as: :taggable,
                      class_name: 'ActsAsTaggableOn::Tagging',
-                     dependent: :destroy
+                     dependent: :destroy,
+                     after_add: :dirty_tag_list,
```
Erowlin: Shall we name the callback `dirtify_tag_list` instead? Same for the `after_remove` callback.

Fodoj:
@Erowlin I've incorporated your valuable feedback :) I've also updated PR description, dropped Rails 4.2 support and fixed tests. They still somehow failed in same combinations, I will try re-triggering to see what happens.

criticerz:
@Fodoj Thanks! Really saved our butts with this PR.

lunaru:
I noticed that there's also another PR for a similar fix: https://github.com/mbleigh/acts-as-taggable-on/pull/862

Any relationship between these two branches? 

Anyway, I'd love to see this get merged!

Fodoj:
@lunaru yes, I used code from 862 in this PR, forgot to mention it

PhilCoggins:
@Fodoj Shouldn't the old version also be tracked?

Example:
```
i.label_list
=> ["foo", "bar", "low priority"]
i.label_list = "foo"
=> "foo"
i.changes
=> {"label_list"=>[nil, ["foo"]]}
```

I would expect the old array to display as the first value in `i.changes` array.

travisofthenorth:
🎉 very happy to see this close to the finish line

cantino:
Is this fully blocked on Rspec 3.8?

spaghetticode:
👍 I'm having issues with rails 5.2 and the current gem release so I hope this gets merged soon

PhilCoggins:
@Fodoj Any thoughts on my comment above? I think this is at least worth addressing before merging. The changeset always being `nil` -> `[foo, bar]` isn't correct.

The reason this isn't working is because the model's [attribute](https://github.com/mbleigh/acts-as-taggable-on/pull/887/files#diff-6fb8d10ba793f1b09e64f2923988ee4eR40) is not being initialized with the current list.

cireficc:
You, my man, are a fricking *life saver*! I'm very excited to see this merged!

Fodoj:
@PhilCoggins which Rails version do you use? There are tests for this behavior and from what I can see tracking works fine, for example:

```ruby
taggable = TaggableModel.create(tag_list: 'awesome, epic')
#=> #<TaggableModel:0x000056219f1b3200 id: 2, name: nil, type: nil>
taggable.tag_list = 'foo'
#=> "foo"
taggable changes
#=> {"tag_list"=>[["awesome", "epic"], ["foo"]]}
```

See also spec/acts_as_taggable_on/taggable/dirty_spec.rb

Basically, them gem already had pretty good test coverage around dirty changes, so I relied on them to tell me if behaviour is the same.

PhilCoggins:
@Fodoj Yes, if the model is initialized with the label list, then it works properly. But what about when I instantiate an already existing record?

```
i = Issue.create(label_list: "foo, bar")
i.label_list = "baz"
i.changes
=> {"label_list"=>[["foo", "bar"], ["baz"]]} # Good
i = Issue.find(i.id)
i.label_list = "baz"
i.changes
=> {"label_list"=>[nil, ["baz"]]} # Bad
```

Using Rails 5.1.4, FWIW.

Fodoj:
Got it now, let me fix it :)

Fodoj:
@PhilCoggins should be fixed now, though let's wait for Travis to confirm. I've added a spec as a proof that it's fixed: https://github.com/mbleigh/acts-as-taggable-on/pull/887/commits/7b4911aa6b84d6bdf4663aa067a5f617a5ee8014.

PhilCoggins:
@Fodoj I just tested on my end, works great! Thank you for addressing this 💯 .

brchristian:
Hi @Fodoj, it looks to me like all of the [Rspec commits](https://github.com/rspec/rspec/commits/master) since 3.7 are just edits to the README. Are you sure we can’t just use Rspec 3.7 instead of the `master` branch?

Then we wouldn’t have to wait to merge. :)

Fodoj:
@brchristian you are absolutely right! Not sure now why did I even use edge versions in the first place. I guess nothing stops @Erowlin from merging this PR now :-)

hardreggaecafe:
When this PR is going to be merged? I can't wait.

Erowlin:
Hey, i just reviewed the code, i have no merge rights on this repository, otherwise it would’ve been merged for a long time ;)

brchristian:
Would @seuros be the person to talk to here? Seems like a few folks (myself included) are waiting on this for Rails 5.2 compatibility.

Fodoj:
if that's won't be merged in next couple of weeks, I will just rename it to acts-as-taggable-on-next and publish it on rubygems so people can install it easily. Agree?

glendel:
This gem should be handled and administrated by the Ruby and Rails community in order to keep it always up-to-date and improved!

patrickoramah:
Hello, I'm really confused as to how to implement this fix in my rails app. If anyone can point me in the right direction that'll be really awesome. Thanks in anticipation.

brchristian:
Hi @patrickoramah until this PR gets merged, you can use the branch where the PR is coming from in your Gemfile, like so:

````ruby
gem 'acts-as-taggable-on', github: 'Fodoj/acts-as-taggable-on', branch: 'rails-5.2'
````

File: Appraisals [hunk 4]
```diff
@@ -1,11 +1,11 @@
+appraise 'activerecord-5.2' do
+  gem 'rails', '5.2.0'
```
seuros: why did we require rails instead of just activerecord ?

acts-as-taggable-on.gemspec
```diff
@@ -22,12 +22,14 @@ Gem::Specification.new do |gem|
...
   end
 
-  gem.add_runtime_dependency 'activerecord', ['>= 4.2.8']
+  gem.add_runtime_dependency 'activerecord', ['>= 5.0']
 
   gem.add_development_dependency 'sqlite3'
   gem.add_development_dependency 'mysql2', '~> 0.3'
   gem.add_development_dependency 'pg'
 
+  gem.add_dependency 'rake'
```
seuros: why did we add this dependency ?

Fodoj:
@seuros shot myself in a leg, but now it's fine, single beautiful commit with all the changes. Also added changes you've requested.

seuros:
@Fodoj Thank you. Do you know why Postgresql build started to fail after  the rebase ?

Fodoj:
@seuros it started to fail because master branch is failing https://github.com/mbleigh/acts-as-taggable-on/commits/master

Appraisals
```diff
@@ -1,11 +1,11 @@
+appraise 'activerecord-5.2' do
+  gem 'activerecord', '5.2.0'
```
Erowlin: Why not using a version like `~> 5.2.0`?

Erowlin: Also, why specifically 5.2 and not 5.0? In the Gemspec you specified 5.0, I suggest we stick with that version.

gemfiles/activerecord_5.2.gemfile
```diff
@@ -0,0 +1,15 @@
+# This file was generated by Appraisal
+
+source "https://rubygems.org"
+
+gem "activerecord", "5.2.0"
```
Erowlin: Same comment, why not sticking with 5.0?

lib/acts_as_taggable_on/taggable/core.rb
```diff
@@ -28,26 +29,43 @@ def initialize_acts_as_taggable_on_core
...
             has_many context_tags, -> { order(taggings_order) },
                      class_name: 'ActsAsTaggableOn::Tag',
                      through: context_taggings,
                      source: :tag
+
+            attribute "#{tags_type.singularize}_list".to_sym, ActiveModel::Type::Value.new
           end
 
           taggable_mixin.class_eval <<-RUBY, __FILE__, __LINE__ + 1
+
```
Erowlin: No need for this new line!

lib/acts_as_taggable_on/taggable/core.rb
```diff
@@ -28,26 +29,43 @@ def initialize_acts_as_taggable_on_core
...
               set_tag_list_on('#{tags_type}', new_tags)
             end
 
             def all_#{tags_type}_list
               all_tags_list_on('#{tags_type}')
             end
+
+            private
+            def dirtify_tag_list tagging
```
Erowlin: You should add parenthesis for the parameters, it seems to be what's done in this project. 
```ruby
def dirtify_tag_list(tagging)
  ...
end
```

lib/acts_as_taggable_on/taggable/core.rb
```diff
@@ -315,3 +316,4 @@ def find_or_create_tags_from_list_with_context(tag_list, _context)
     end
   end
 end
+
```
Erowlin: 👍

lib/acts_as_taggable_on/taggable/tagged_with_query/query_base.rb
```diff
@@ -48,7 +48,7 @@ def tags_match_type
 
     def escaped_tag(tag)
       tag = tag.downcase unless ActsAsTaggableOn.strict_case_match
-      ActsAsTaggableOn::Utils.escape_like(tag)
+      tag.gsub(/[!%_]/) { |x| '!' + x }
```
Erowlin: This one should be crazily spec'ed :).

spec/acts_as_taggable_on/dirty_spec.rb
```diff
@@ -14,19 +14,27 @@
       end
 
       it 'should show changes of dirty object' do
-        expect(@taggable.changes).to eq({'tag_list' => ['awesome, epic', ['one']]})
+        expect(@taggable.changes).to eq({'tag_list' => [['awesome', 'epic'], ['one']]})
```
Erowlin: ⚠️ 

This is a breaking change. 

Either rollback your change or we have to communicate crazy about that. Our own internal code depends on the previous way it was formatted, and I bet a bunch of projects have the same.

Fodoj: @Erowlin this PR is tagged for release 6.0.0, which is supposed to have breaking changes, no?

Erowlin: Then I guess we're fine!

lib/acts_as_taggable_on/taggable/tagged_with_query/query_base.rb
Fodoj: It's just a rollback of https://github.com/mbleigh/acts-as-taggable-on/commit/cc7a2c097ee9eeb4417328986f0dd49112cb05b3, nothing I wrote for this PR. That commit breaks builds in master branch.

Appraisals
Fodoj: I don't get that. This is PR to add support for 5.2, why would I specify 5.0 in 5.2 appraisal?

Erowlin: Well, we're dropping support for 4.2, it does not mean that we drop support for 5.0 I guess, correct?

Fodoj:
I pushed more fixes. I actually assumed that after 3.5 months someone familiar with the code base already reviewed it, but seems like it's not the case?..

lib/acts_as_taggable_on/taggable/core.rb
Erowlin: This one was fine actually, It's good to have a blank line at the end of files :). Sorry, I should have been more specific than a 👍  
https://stackoverflow.com/questions/2287967/why-is-it-recommended-to-have-empty-line-in-the-end-of-file

Erowlin:
I am not very familiar with the acts-as-taggable codebase unfortunately.

Fodoj:
I assume as @seuros has permissions to merge to master, he is familiar. Let's wait for his final go and merge then.

Added new line back also.

Appraisals
Fodoj: Correct, that's why Appraisal file has definitions for 5.0, 5.1 and now for 5.2

lib/acts_as_taggable_on/taggable/core.rb
```diff
@@ -28,12 +29,16 @@ def initialize_acts_as_taggable_on_core
...
+                     dependent: :destroy,
+                     after_add: :dirtify_tag_list,
+                     after_remove: :dirtify_tag_list
 
             has_many context_tags, -> { order(taggings_order) },
                      class_name: 'ActsAsTaggableOn::Tag',
                      through: context_taggings,
                      source: :tag
+
+            attribute "#{tags_type.singularize}_list".to_sym, ActiveModel::Type::Value.new
```
ebihara99999: This line makes `taggable_obj.attribute_names` changed. Specs failed in the following case, using `json.extract! @item, *@item.attribute_names` as response. 

I saw you comment this PR will be released as the version 6.0.0 and allow breaking changes, but the impact of this change, if you don't expect, is very big.

What do you think about this case?

ebihara99999: @Fodoj

Fodoj: Not sure I understand the issue. Could you provide a failing spec?

Though I think if it's 6.0.0 release, then breaking changes are totally ok, just repo owner has to update readme before reasling this new version.

ebihara99999: Failing is request spec with the [rspec-json_matcher gem](https://github.com/r7kamura/rspec-json_matcher). I copy and modify some from the real codes and output of running rspec:

```ruby
# request spec
it { is_expected.to be_json_as [product: expected_product_response] }
```

```
# output
expected to match:

  {
    :product => {
      :id          => Integer < Numeric,
      :user_id     => Integer < Numeric,
      :product_id  => Integer < Numeric,
      :created_at  => String < Object,
      :updated_at  => String < Object,
    }
  }

actual:
  {
    :product => {
      "id"          => 2,
      "user_id"     => 11,
      "product_id"  => 216,
      "tag_list"    => []
      "created_at"  => "2018-06-05T17:36:33+09:00",
      "updated_at"  => "2018-06-05T17:36:33+09:00"
    }
  }


reason: product "tag_list"
```

ebihara99999: The case mentioned above is with the rspec-json_matcher gem, but if not using the rspec-json_matcher gem, the test suites testing api response would all fail because of the change.

lib/acts_as_taggable_on/taggable/collection.rb
```diff
@@ -174,7 +174,7 @@ module CalculationMethods
       # See https://github.com/rails/rails/blob/master/activerecord/lib/active_record/relation/calculations.rb#L38
       def count(column_name = :all, options = {})
```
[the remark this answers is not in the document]
Fodoj: I tried - some of tests fail, especially sqlite ones. Let's not risk it this time :)

spec/acts_as_taggable_on/dirty_spec.rb
seuros: I think this is a major breaking change that we should avoid. 
People expect to have tags in the root of the array not inside another array. 
Can we try to avoid this ?

Fodoj: I can not promise I will have any time to invest into it in next days/weeks. Even though new behaviour seems much more logical to me, because previously it was a change from string to array, and now it's change from array to array ( no akward change of data type).

lib/acts_as_taggable_on/taggable/tagged_with_query/query_base.rb
Fodoj: Not to repeat myself, but this change is unrelated to the scope of this PR, as it barely rollbacks commit that broke the build here and in master branch

Onikoroshi:
Any idea when this will be available as a version? I'm still only able to get version 5.0.0 which doesn't seem to include this fix ...

rsl:
@Fodoj i'm trying to work on "fixing" the PR that supposedly breaks the build. looking at the Travis report: https://travis-ci.org/mbleigh/acts-as-taggable-on/jobs/339031800#L657 it seems to be a gem/config issue that i cannot replicate nor see how my changes could possibly have affected. 

`Specified 'postgresql' for database adapter, but the gem is not loaded. Add `gem 'pg'` to your Gemfile (and ensure its version is at the minimum required by ActiveRecord).` 

i cannot get travis to retry my fork because the PR was closed.

lib/acts_as_taggable_on/taggable/core.rb
akostadinov: In my opinion this change is harmful #1064

PR #889, opened 2018-03-12
comply with Evia's changes

eitangenis:


seuros:
This change seem unrelated to this gem, why we have account.id ?

PR #893, opened 2018-04-27
Simple reminder to save to persist `tag_list`

raelgc:
Just a simple reminder. In a quick read I had the impression that calling `add` and `remove` for `tag_list` was automatically persisting data.


seuros:
Thank you

PR #894, opened 2018-05-04
Fix ActiveRecord::StatementInvalid error when using UUID as primary key

CalvertYang:
If using UUID as primary key, the relationships feature raises `ActiveRecord::StatementInvalid` error when you try something like:

```
User.first.find_related_skills
```

```
ActiveRecord::StatementInvalid: PG::SyntaxError: ERROR:  syntax error at or near "a56f2e"
LINE 1: ...w FROM users, tags, taggings WHERE (users.id != 96a56f2e-393...
                                                             ^
```

### DIFF
--- a/lib/acts_as_taggable_on/taggable/related.rb
+++ b/lib/acts_as_taggable_on/taggable/related.rb
@@ -49,7 +49,7 @@ module ActsAsTaggableOn::Taggable
     private
 
     def exclude_self(klass, id)
-      "#{klass.table_name}.#{klass.primary_key} != #{id} AND" if [self.class.base_class, self.class].include? klass
+      "#{klass.arel_table[klass.primary_key].not_eq(id).to_sql} AND" if [self.class.base_class, self.class].include? klass
     end
 
     def group_columns(klass)
--- a/spec/acts_as_taggable_on/related_spec.rb
+++ b/spec/acts_as_taggable_on/related_spec.rb
@@ -33,6 +33,15 @@ describe 'Acts As Taggable On' do
       expect(taggable1.find_related_tags).to_not include(taggable2)
     end
 
+    it 'should find related objects based on tag names on context - uuid primary key' do
+      taggable1 = TaggableModelWithUuidPrimaryKey.create!(name: 'Taggable 1',tag_list: 'one, two')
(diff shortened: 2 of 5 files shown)


lib/acts_as_taggable_on/taggable/related.rb
```diff
@@ -49,7 +49,7 @@ def related_tags_for(context, klass, options = {})
     private
 
     def exclude_self(klass, id)
-      "#{klass.table_name}.#{klass.primary_key} != #{id} AND" if [self.class.base_class, self.class].include? klass
+      "#{klass.table_name}.#{klass.primary_key} != '#{id}' AND" if [self.class.base_class, self.class].include? klass
```
seuros: This fixed uuid PK but broke integer/bigint

CalvertYang: @seuros Thank you for pointing out my mistake.

How do you think if I use `klass.columns_hash['id'].type` to check the PK type first, and modify the method like this:

```rb
def exclude_self(klass, id)
  new_id = klass.columns_hash['id'].type == :uuid ? "'#{id}'" : id

  "#{klass.table_name}.#{klass.primary_key} != #{new_id} AND" if [self.class.base_class, self.class].include? klass
end
```

seuros: I think we should use Arel in this part of code.

CalvertYang: @seuros Thanks for your suggestion.
I'm not familiar with Arel, would you please confirm the following code is correct.

```rb
def exclude_self(klass, id)
  "#{klass.arel_table[klass.primary_key].not_eq(id).to_sql} AND" if [self.class.base_class, self.class].include? klass
end
```

seuros: Yes, i think it should work.
Can you rebase your PR ?

seuros: Also please include a test case.

seuros:
Thank you
We could generate uuid in postgresql and leave integer for mysql and sqlite3

CalvertYang:
Oh, I see. But I have no idea about how to modify the `spec/internal/db/schema.rb` that only generate `uuid` in postgresql and leave `integer` for mysql and sqlite3.

seuros:
cherry-picked on #898 .

PR #901, opened 2018-06-19
fix camelCase column lead to exception when query

seaify:
when the pg database have column, like `firstName`, the original code would throw exception.

need add quote for the column names


seuros:
I seem we need to use Arel for this part

PR #906, opened 2018-07-09
Use none instead of false query

gssbzn:
None allows the use a null object pattern, this prevent doing further queries to the database knowing they'll always return false

Originally
```ruby
posts = Post.tagged_with.recent
```
Would still generate queries to the database
with this change Rails would always return a chainable NullObject that avoids doing further queries tothe database


gssbzn:
@seuros updated 👍

PR #911, opened 2018-07-24
Fix User-Defined Column Cache Override

tonyta:
#### Problem

We ran into a problem where this gem was unexpectedly overriding our defined `columns` method. This was because this gem uses `instance_eval` which defines methods directly on the class. 

#### Before

Given the following schema and model, the `columns` method should reject the `ignored_column`:
```ruby
create_table :columns_override_models, force: true do |t|
  t.column :name, :string
  t.column :type, :string
  t.column :ignored_column, :string
end
```

```ruby
class ColumnsOverrideModel < ActiveRecord::Base
  def self.columns
    super.reject { |c| c.name == 'ignored_column' }
  end
end

ColumnsOverrideModel.columns.map(&:name)
# => ["id", "name", "type"]
```

Unfortunately, invoking `acts_as_taggable` on the model after the user has defined `columns` will override the user's method using `instance_eval`.
```ruby
class ColumnsOverrideModel < ActiveRecord::Base
  def self.columns
    super.reject { |c| c.name == 'ignored_column' }
  end
end

ColumnsOverrideModel.columns.map(&:name)
# => ["id", "name", "type", "ignored_column"] 😳
```

#### Solution

The solution in this PR is to user `extend` instead of `instance_eval`, which will not override the user-defined method, and will define the intercept method as part of the inheritance ancestor walk-up.
```ruby
class ColumnsOverrideModel < ActiveRecord::Base
  def self.columns
    super.reject { |c| c.name == 'ignored_column' }
  end
end

ColumnsOverrideModel.columns.map(&:name)
# => ["id", "name", "type"] 😊
```

---

In addition, since I was confused about where to defined my test models, this PR also cleans up the definition and loading of those models. Hope this is helpful! 💜❤️🧡💛💚💙


tonyta:
Thanks for your patience, @seuros. I've added the changelog entry and rebased. 🤘

PR #923, opened 2018-09-29
Scope tags for taggable type

florianeck:
just realized when suggesting autocomplete for tags to users, i cant show tags only for a certain `taggable_type`
(maybe i missed it in the code though) - if not here is a patch for that situation.

in may case - tags are totally different depending on the class the are assigned to. now i can filter for that as well.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -2,6 +2,10 @@ source 'https://rubygems.org'
 
 gemspec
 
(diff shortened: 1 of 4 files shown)


seuros:
This new scope will need a Readme and Changelog entry.

PR #925, opened 2018-10-07
Set belongs_to as optional to get around new Rails 5.2 presence enforcement

coneybeare:
Without this simple change, we get problems with creating tags as part of a record creation, rather than adding them afterwards.

### DIFF
--- a/lib/acts_as_taggable_on/tagging.rb
+++ b/lib/acts_as_taggable_on/tagging.rb
@@ -4,7 +4,7 @@ module ActsAsTaggableOn
 
     DEFAULT_CONTEXT = 'tags'
     belongs_to :tag, class_name: '::ActsAsTaggableOn::Tag', counter_cache: ActsAsTaggableOn.tags_counter
-    belongs_to :taggable, polymorphic: true
+    belongs_to :taggable, polymorphic: true, optional: true
 
     belongs_to :tagger, {polymorphic: true}.tap {|o| o.merge!(optional: true) }
 


coneybeare:
Only error was a travis-ci setup timeout:

````
travis_fold:start:worker_info
[0K[33;1mWorker information[0m
hostname: 0e9443f5-253f-4b27-9f6a-13eec83d971d@1.production-1-worker-org-01-packet
version: v4.1.2 https://github.com/travis-ci/worker/tree/91246b057ccd93649046771f29221f839c19a7d3
instance: dfe81bd travisci/ci-garnet:packer-1512502276-986baf0 (via amqp)
startup: 2.887309412s
travis_fold:end:worker_info
[0KOCI runtime exec failed: exec failed: container_linux.go:348: starting container process caused "process_linux.go:86: executing setns process caused \"exit status 21\"": unknown
````

**Comment by mkllnk** [comment]
I had a similar problem when cloning a record with tags. We did:
```rb
  thing = other_thing.dup
  thing.tag_ids = other_thing.tag_ids
  thing.save!
```

But in the end I found that I could remove the copying of ids. `other_thing.dup` already copies the `tag_list` attribute and everything gets saved correctly. So the new working code is just:

```rb
  thing = other_thing.dup
  thing.save!
```

Technically, a Tagging needs a Taggable. The current code is correct and your patch would make it optional even though it's required. I reckon though, that there may be another bug preventing the validation to succeed even though the Tagging has a valid Taggable (just unsaved, maybe).

PR #928, opened 2018-11-20
Fix caching on first save for an unused model

kratob:
On certain code paths, saving an instance of a model for the first time can cause the cache column not to be written. Affected code looks like this:

```
class Model < ActiveRecord::Base
  acts_as_taggable_on
end

record = Model.new
record.tag_list = 'foo, 'bar'
record.save!
record.tag_list   # => nil
```

The reason for this is that the `after_save` hook gets dynamically added when calling `Model.columns`. However, `Model.new` does not actually do that.

The fix uses `Model.load_schema!` instead.


judithroth:
This would be really useful to us. Is there something that blocks merging this (apart from the conflicts that are now there because the MR is quite old)? 
We would happily update the MR if there is a chance that this gets merged.

FLeinzi:
Is there any chance to bring this PR up-to-date?

seuros:
The caching code this patched was rewritten in v12 and the file it touches no longer exists. The save hooks are registered up front now, so the first-save scenario this fixed cannot happen anymore. Thanks for the fix and the spec, and sorry it sat this long.

PR #930, opened 2018-12-03
use ActiveRecord::Migration version 5.0 or 5.1 instead of 4.2

wayne5540:
## Why

1. Since this gem is dropping support for Rails 4.2 (#887), there is no reason to use `ActiveRecord::Migration[4.2]` as default migration version, sooner we migrate to 5 better
2. In Rails version > 5.1, default type of primary key becomes `bigint`, use `ActiveRecord::Migration[5.1]` can also help us adapt this change, make it more consistent

## Notable point

1. In Rails version > 5, migration method `t.references` will add index automatically, to keep migration result consistent, I add `index: false` for `tag`, `taggable `, `tagger` to prevent extra `index_taggings_on_taggable_type_and_taggable_id` and `index_taggings_on_tagger_type_and_tagger_id` index
2. After this change, if your ActiveRecord version > 5.1, all table's id will become `bigint` (if your database supports `bigint`)

### DIFF
--- a/db/migrate/1_acts_as_taggable_on_migration.rb
+++ b/db/migrate/1_acts_as_taggable_on_migration.rb
@@ -1,7 +1,7 @@
-if ActiveRecord.gem_version >= Gem::Version.new('5.0')
-  class ActsAsTaggableOnMigration < ActiveRecord::Migration[4.2]; end
+if ActiveRecord.gem_version >= Gem::Version.new('5.1')
+  class ActsAsTaggableOnMigration < ActiveRecord::Migration[5.1]; end
 else
-  class ActsAsTaggableOnMigration < ActiveRecord::Migration; end
+  class ActsAsTaggableOnMigration < ActiveRecord::Migration[5.0]; end
 end
 ActsAsTaggableOnMigration.class_eval do
   def self.up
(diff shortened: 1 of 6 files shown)


nbulaj:
Why not just use `[#{ActiveRecord::VERSION::MAJOR}.#{ActiveRecord::VERSION::MINOR}]` to support any AR version?

wayne5540:
@nbulaj Yes, agree with you, since `.gemspec` now has run time `activerecord` 
dependency `>= 5.0, < 6.1`, we probably don't need the `if else` statement anymore and we can just use `[#{ActiveRecord::VERSION::MAJOR}.#{ActiveRecord::VERSION::MINOR}]`, thanks for pointing this out.

I'm happy to do the change, however, this PR has been idle here for a while, I'm not sure if this change is wanted 😅

nbulaj:
> I'm happy to do the change, however, this PR has been idle here for a while, I'm not sure if this change is wanted sweat_smile

You need to ask them, try to ping @seuros :)

seuros:
@wayne5540  Thank you for your contribution, can you rebase this PR and have it use version 6.0 instead.

wayne5540:
Looks like all the files has been updated to 6.0 version, nice work team! This PR is no longer needed.

PR #931, opened 2018-12-06
Add a test case for previous_changes

rspeicher:
This is attempting to reproduce an odd behavior we're seeing in GitLab at https://gitlab.com/gitlab-org/gitlab-ee/merge_requests/5107#note_122281994.

The problem we're seeing may be an odd interaction between ActiveRecord and some other gem we're using, but I want to first verify it's not an issue with acts-as-taggable-on itself.

### DIFF
--- a/spec/acts_as_taggable_on/dirty_spec.rb
+++ b/spec/acts_as_taggable_on/dirty_spec.rb
@@ -37,6 +37,14 @@ describe 'Dirty behavior of taggable objects' do
         expect(@taggable.tag_list_change).to eq([['awesome', 'epic'], ['one']])
       end
 
+      it 'populates previous_changes correctly' do
+        @taggable.save!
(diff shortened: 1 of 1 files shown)


spec/acts_as_taggable_on/dirty_spec.rb
```diff
@@ -37,6 +37,14 @@
         expect(@taggable.tag_list_change).to eq([['awesome', 'epic'], ['one']])
       end
 
+      it 'populates previous_changes correctly' do
+        @taggable.save!
+        @taggable.update!(tag_list: %w(two three four))
```
stanhu: The test fails if I add `@taggable.reload` here.

seuros:
Closing this one. Dirty tracking was rebuilt on top of the ActiveModel attribute API since this was opened, so the behavior this spec was probing no longer maps to the current code. Thanks anyway.

PR #937, opened 2019-01-19
Rails 6.0.0.beta1 support

damianlegawiec:


seuros:
Thank you @damianlegawiec , i will release a prerelease version on Monday.

PR #951, opened 2019-05-02
Rails 6 fixes

damianlegawiec:


igorkasyanchuk:
with Rails 6 it doesn't work, see travis.ci logs.

damianlegawiec:
@igorkasyanchuk fixed!

igorkasyanchuk:
great!

dankmitchell:
Nice work!

seuros:
We need to poing out in the changelog before pushing a new version out.
I see that tag_list changed it format in rails 6

behindthebid:
+1

gkubaryk:
can we also update the validator in `lib/acts_as_taggable_on/tag.rb` as such:
```
-    validates_uniqueness_of :name, if: :validates_name_uniqueness?
+    validates_uniqueness_of :name, if: :validates_name_uniqueness?, case_sensitive: true
```
This will address the deprecation warning:
DEPRECATION WARNING: Uniqueness validator will no longer enforce case sensitive comparison in Rails 6.1. To continue case sensitive comparison on the :name attribute in ActsAsTaggableOn::Tag model, pass `case_sensitive: true` option explicitly to the uniqueness validator.

seuros:
@gkubaryk i don't think tags should be case_sensitive.

gkubaryk:
@seuros i have no strong opinion either way; the important thing (to me) is to set it to *something* so that the behavior is defined moving forward. the deprecation warning implied to me that it's silently been acting case sensitive and no longer will in the future. if you have a different reading of it or if we should take the opportunity to explicitly set it false, no argument here. :)

taf2:
Pretty sure the default is case sensitive false

seuros:
@taf2  , When you use `case_sensitive : false`, the column is wrapped with `lower()` function in both side to remove sensitivity.
You dont need it if your field is already case insensitive eq [citext](https://www.postgresql.org/docs/current/citext.html).

spec/acts_as_taggable_on/dirty_spec.rb
```diff
@@ -20,7 +20,11 @@
       it 'should show changes of freshly initialized dirty object' do
         taggable = TaggableModel.find(@taggable.id)
         taggable.tag_list = 'one'
-        expect(taggable.changes).to eq({'tag_list' => [['awesome', 'epic'], ['one']]})
+        if Rails.version >= "6.0.0.beta1"
```
murb: I don't think this is the right call; I have tried fixing this but I'd suggest starting this it block with 

`pending if Rails.version >= "6.0.0.beta1"`

and remove the conditional. Here you're silently allowing a possibly breaking change to simply pass the tests

lib/acts_as_taggable_on/taggable/core.rb
```diff
@@ -50,7 +50,12 @@ def #{tag_type}_list=(new_tags)
               parsed_new_list = ActsAsTaggableOn.default_parser.new(new_tags).parse
 
               if self.class.preserve_tag_order? || parsed_new_list.sort != #{tag_type}_list.sort
-                set_attribute_was('#{tag_type}_list', #{tag_type}_list)
+                if Rails.version < "6.0.0.beta1"
+                  set_attribute_was('#{tag_type}_list', #{tag_type}_list)
+                else
+                  duplicated_mutations_from_database.change_to_attribute('#{tag_type}_list')
```
murb: Are you sure this is even working? I get the same result without it.

lib/acts_as_taggable_on/taggable/core.rb
```diff
@@ -50,7 +50,12 @@ def #{tag_type}_list=(new_tags)
               parsed_new_list = ActsAsTaggableOn.default_parser.new(new_tags).parse
 
               if self.class.preserve_tag_order? || parsed_new_list.sort != #{tag_type}_list.sort
-                set_attribute_was('#{tag_type}_list', #{tag_type}_list)
+                if Rails.version < "6.0.0.beta1"
```
ogasawaraShinnosuke: Does not work in Rails 6.0.0 environment.
https://github.com/ogasawaraShinnosuke/acts-as-taggable-on/commit/aa602f319f85ab32ac63403b6863cf4be3c4d6a2

tbuehl: `if Rails::MAJOR::VERSION < 6`

`Rails.version` is just a string.

```
"6.0.0" < "6.0.0.beta1"
=> true
```

which is not what you want

tbuehl:
For anyone on Rails 6 and as long as this PR isn't updated, here is a working solution:

https://github.com/tbuehl/acts-as-taggable-on/tree/fix/rails-6-and-failing-specs

edd1312:
I changed the **core.rb** file to:

```
if self.class.preserve_tag_order? || parsed_new_list.sort != #{tag_type}_list.sort
                # set_attribute_was('#{tag_type}_list', #{tag_type}_list)
                write_attribute("#{tag_type}_list", #{tag_type}_list)
                attribute_will_change! "#{tag_type}_list"
                save
                write_attribute("#{tag_type}_list", parsed_new_list)
              end
```

and it works !!!

wpliao1989:
> can we also update the validator in `lib/acts_as_taggable_on/tag.rb` as such:
> 
> ```
> -    validates_uniqueness_of :name, if: :validates_name_uniqueness?
> +    validates_uniqueness_of :name, if: :validates_name_uniqueness?, case_sensitive: true
> ```
> 
> This will address the deprecation warning:
> DEPRECATION WARNING: Uniqueness validator will no longer enforce case sensitive comparison in Rails 6.1. To continue case sensitive comparison on the :name attribute in ActsAsTaggableOn::Tag model, pass `case_sensitive: true` option explicitly to the uniqueness validator.

I'm wondering if there's an official fix for this issue? This is generating a lot of deprecation warnings.

curtp:
We're using gem version 6.5.0 and we're seeing a lot of the deprecation warnings about the uniqueness validator.

fnguyen-tc:
Please change to `validates_uniqueness_of :name, if: :validates_name_uniqueness?, case_sensitive: false`

PR #962, opened 2019-08-25
Fix `Deprecate Unicode#downcase/upcase/swapcase.`

sin-to-jin:
```rb
DEPRECATION WARNING: ActiveSupport::Multibyte::Unicode#downcase is deprecated and will be removed from Rails 6.1. Use String methods directly. (called from unicode_downcase at /vendor/bundle/ruby/2.6.0/bundler/gems/acts-as-taggable-on-d01c315f616b/lib/acts_as_taggable_on/tag.rb:126)
```

ref rails/rails@619b2ae

### DIFF
--- a/lib/acts_as_taggable_on/taggable/core.rb
+++ b/lib/acts_as_taggable_on/taggable/core.rb
@@ -53,7 +53,7 @@ module ActsAsTaggableOn::Taggable
                 if Rails.version.to_f < 6.0
                   set_attribute_was('#{tag_type}_list', #{tag_type}_list)
                 else
-                  duplicated_mutations_from_database.change_to_attribute('#{tag_type}_list')
+                  d_mutations_from_database.change_to_attribute('#{tag_type}_list')
                 end
                 write_attribute("#{tag_type}_list", parsed_new_list)
               end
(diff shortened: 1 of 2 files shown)


lib/acts_as_taggable_on/taggable/core.rb
```diff
@@ -65,6 +69,17 @@ def all_#{tags_type}_list
...
+            def duplicated_mutations_from_database
+            unless defined?(@mutations_from_database)
+              @mutations_from_database = nil
+            end
+            @mutations_from_database ||= if defined?(@attributes)
+              ActiveModel::AttributeMutationTracker.new(@attributes)
+            else
+              NullMutationTracker.instance
+            end
+          end
```
sunny: Indentation seems off for this method.

lib/acts_as_taggable_on/taggable/core.rb
```diff
@@ -65,6 +69,17 @@ def all_#{tags_type}_list
             def dirtify_tag_list(tagging)
               attribute_will_change! tagging.context.singularize+"_list"
             end
+              
+            def duplicated_mutations_from_database
+            unless defined?(@mutations_from_database)
+              @mutations_from_database = nil
+            end
```
sunny: I don't think you need these three lines. `@mutations_from_database ||=` should be enough to initialize it to `nil`.

lib/acts_as_taggable_on/taggable/core.rb
ogasawaraShinnosuke: Reverted because there was an extra commit.

lib/acts_as_taggable_on/taggable/core.rb
ogasawaraShinnosuke: Reverted because there was an extra commit.

ogasawaraShinnosuke:
Reverted because there was an extra commit.

PR #967, opened 2019-09-08
Support Rails 6.0.0

john:
This PR consists of removing ".beta1" from the gemfile, so it will install on Rails 6.0.0


seuros:
thanks, but there is still one failing test to work with AR 6.0. Check #968

PR #1005, opened 2020-07-27
Add case sensitivity on Tag name for Rails 6

andoke:
I'm seeing lots of deprecations warnings in Rails 6.
This fixes the issue.

### DIFF
--- a/lib/acts_as_taggable_on/tag.rb
+++ b/lib/acts_as_taggable_on/tag.rb
@@ -10,7 +10,7 @@ module ActsAsTaggableOn
     ### VALIDATIONS:
 
     validates_presence_of :name
-    validates_uniqueness_of :name, if: :validates_name_uniqueness?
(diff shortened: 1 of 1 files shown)


lib/acts_as_taggable_on/tag.rb
```diff
@@ -10,7 +10,7 @@ class Tag < ::ActiveRecord::Base
     ### VALIDATIONS:
 
     validates_presence_of :name
-    validates_uniqueness_of :name, if: :validates_name_uniqueness?
+    validates_uniqueness_of :name, if: :validates_name_uniqueness?, case_sensitive: true
```
graaff, line 13: This would have to use ActsAsTaggableOn.strict_case_match rather than a hard-coded true.

andoke:
Closing this, #1009 is a better solution.

PR #1009, opened 2020-10-28
Use a custom validation method to validate the uniqueness of name

FunnyHector:
Use a custom validation method to validate the uniqueness of name

The intention of this change is to address the the following deprecation seen on Rails 6:

    DEPRECATION WARNING: Uniqueness validator will no longer enforce
    case sensitive comparison in Rails 6.1. To continue case sensitive
    comparison on the :name attribute in ActsAsTaggableOn::Tag model,
    pass `case_sensitive: true` option explicitly to the uniqueness
    validator.

Since there is the complexity of the config `ActsAsTaggableOn.strict_case_match`, we can't just explicitly set case_sensitive option as suggested in the deprecation warning.

The tests are quite blargh to deal with since the test database has the collation set to case-insensitive and has an uniqueness index, which is not realistic. To deal with the tests, I used a similar approach as in cb8d6e65cbb65a02a05f3deb28c12d5439d745af. In cases where case sensitivity matters, we alter the collation on `name` column, run the test, then alter the collation back.

This PR tries to solve the same problem mentioned in #1005, but since the author of #1005 hasn't responded, I'm trying my luck here.

This change has also adopted [@graaff's suggestion](https://github.com/mbleigh/acts-as-taggable-on/pull/1005#discussion_r511326975)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,10 @@ Each change should fall into categories that would affect whether the release is
 
 As such, _Breaking Changes_ are major. _Features_ would map to either major or minor. _Fixes_, _Performance_, and _Misc_ are either minor or patch, the difference being kind of fuzzy for the purposes of history. Adding _Documentation_ (including tests) would be patch level.
 
+### [Master / Unreleased](https://github.com/mbleigh/acts-as-taggable-on/compare/v9.0.1...master)
+* Fixes
+  * [@FunnyHector Fix the uniqueness validator deprecation warning on Rails 6](https://github.com/mbleigh/acts-as-taggable-on/pull/1009)
+
 ### [v9.0.1) / 2022-01-07](https://github.com/mbleigh/acts-as-taggable-on/compare/v8.1.0...v9.0.0)
 * Fixes
   * Fix migration that generate default index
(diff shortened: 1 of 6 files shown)


FunnyHector:
Any idea if this would get noticed soon?

FunnyHector:
Any idea if this would get noticed soon?

seuros:
@FunnyHector , can you rebase this PR. It good to merge.

FunnyHector:
Hi @seuros , I have rebased this change onto the latest master. There was some conflict with [this line in #1013](https://github.com/mbleigh/acts-as-taggable-on/pull/1013/files#diff-ddec22168a27b3e2225ae3567266fd01cd9baf4ed7e9f2a3c9c95b0afcc89b2cR13). 

Could you please have another look?

Sorry for late response. Somehow I must have missed the ping.

alto:
@FunnyHector Do you mind merging `master` back into the PR again?

seuros:
@alto  the only issue left here is the non i18n error message.

alto:
@seuros Yeah, the one in `name_is_unique`. I agree. 👍🏻  /cc @FunnyHector

FunnyHector:
Hiya, sorry I haven't really worked on any project that used i18n :p, that often slips my mind.

I'll try to make a fix to use i18n, and rebase this work. I'm a little busy this week but hopefully I can get it done by the end of this week. Will ping you when I'm done.

FunnyHector:
@seuros @alto I've rebased this work to resolve the conflicts. I've also added a new commit to use I18n error message. Please have another look. Sorry I haven't found time to look at this until now.

PR #1012, opened 2020-12-10
Rails 6.1 support

jamesbrooks:


kvokka:
This branch does **not** test against rails 6.1 in CI, see #1013

lib/acts-as-taggable-on.rb
```diff
@@ -1,6 +1,7 @@
 require 'active_record'
 require 'active_record/version'
 require 'active_support/core_ext/module'
+require 'active_support/core_ext/string/starts_ends_with'
```
silva96, line 4: why is this needed?

jamesbrooks: Replacing `ends_with?` with `end_with?` would had also been an alternative, that require was to bring in that alias.

jamesbrooks:
@kvokka thanks, addressed in 786bd25

jamesbrooks:
Closing in favor of #1013 which resolves additional issues

PR #1015, opened 2020-12-11
Add support for Rails 6.1

rhymes:
Tentative support for Rails 6.1


seuros:
Already merged a working version in master. Thanks for your attempt @rhymes

PR #1030, opened 2021-04-09
Use none? instead of count.zero?

gr-eg:
We have run in to a performance issue when removing unused tags and not using the `tags_counter`

`taggings.count.zero?` executes the following sql
```sql
`SELECT COUNT(*) FROM taggings WHERE taggings.tag_id = ?`
```

This can be very slow on large tables 

by using `taggings.none?` it instead executes:
```sql
`SELECT 1 AS one FROM taggings WHERE taggings.tag_id = 304185392 LIMIT 1`
```

which can always use the index and is much faster.


seuros:
released

PR #1034, opened 2021-05-10
Added strict_default_collation_match as a config option

JosephLing:
If you don't need tags to be case-sensitive and want to use database indexing instead of using `LOWER`.


seuros (approved):

seuros:
@JosephLing  Seem like some versions are failing the spec.

JosephLing:
Thank you I am back at work now so should be able to hopefully fix them next week. It looks like if I want to run all the tests on all the db types easily I should use `wwtd` ?

PR #1037, opened 2021-06-17
fix #1036 scope taggable#tenant method name to #taggable_tenant

### LINKED ISSUE #1036: Models won't save if they already have a `belongs_to :tenant` and `acts_as_taggable_tenant` is used
# How to reproduce:

```ruby

class Tenant < ApplicationRecord
  has_many :users
end

class User < ApplicationRecord
  belongs_to :tenant
  acts_as_taggable_on :tags
  acts_as_taggable_tenant :tenant_id
end

Tenant.create!
```

# Expected behavior

The tenant is created

# Actual behavior

The following error is thrown.

```
NoMethodError - undefined method `marked_for_destruction?' for ID:Integer
```

# Investigation

This is due to: https://github.com/mbleigh/acts-as-taggable-on/pull/1000/files#diff-f3570227aa94bb5887e3cb70c6a69fb30861dbecaa1ee498153371a1e10fc314R217

This line overrides the tenant method and returns an integer instead of an object.
This can be easily fixed by namescoping the method (e.g. `taggable_tenant_id`).

cc @lunaru (BTW, thank you for your work 👍 )

ngouy:
If your taggable model already have a #tenant attribute/method/relationship, it would broke your code.

A small modification is required to scope this (non-documented and normally non directly used) method to the taggable lexical scope to fix the issue

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -7,5 +7,6 @@ group :local_development do
   gem 'guard-rspec'
   gem 'appraisal'
   gem 'rake'
+  gem 'sqlite3'
   gem 'byebug', platforms: [:mri]
 end
--- a/lib/acts_as_taggable_on/taggable/core.rb
+++ b/lib/acts_as_taggable_on/taggable/core.rb
@@ -214,9 +214,9 @@ module ActsAsTaggableOn::Taggable
       self.class.tag_types.map(&:to_s) + custom_contexts
     end
 
-    def tenant
+    def taggable_tenant
       if self.class.tenant_column
-        read_attribute(self.class.tenant_column)
+        public_send(self.class.tenant_column)
       end
     end
 
@@ -278,8 +278,8 @@ module ActsAsTaggableOn::Taggable
 
         # Create new taggings:
         new_tags.each do |tag|
-          if tenant
-            taggings.create!(tag_id: tag.id, context: context.to_s, taggable: self, tenant: tenant)
+          if taggable_tenant
+            taggings.create!(tag_id: tag.id, context: context.to_s, taggable: self, tenant: taggable_tenant)
           else
             taggings.create!(tag_id: tag.id, context: context.to_s, taggable: self)
           end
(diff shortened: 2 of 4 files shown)


Gemfile
```diff
@@ -7,5 +7,6 @@ group :local_development do
   gem 'guard-rspec'
   gem 'appraisal'
   gem 'rake'
+  gem 'sqlite3'
```
ngouy, line 10: I added that because it's a requirement when working on the gem

without it:
```
>$ bundle
...
>$ bundle exec rake rspec
```

image

ngouy:
tested on a real project with already existing tenants and it works (from breaking the app to working app)

lib/acts_as_taggable_on/taggable/core.rb
```diff
@@ -214,7 +214,7 @@ def tagging_contexts
       self.class.tag_types.map(&:to_s) + custom_contexts
     end
 
-    def tenant
+    def taggable_tenant
       if self.class.tenant_column
         read_attribute(self.class.tenant_column)
```
ngouy, line 219: @lunaru small question: why did we specificly use `#read_attribute`?

If I have a taggable resource that is:
```rb
class MyTaggableModel
  has_one: :tenant, through: :something
  
  acts_as_taggable_on :tags
  acts_as_taggable_tenant :tenant_id
  
  def tenant_id
    something.tenant_id
  end
end
```

It just doesn't works, and that's kinda unfortunate.

Gemfile
seuros: Strange , i never needed to add it. You can use appraisal to run a specific version.

seuros:
Can you bump version 8.1 and update the Changelog ?

ngouy:
@seuros can we also address the #read_attribute issue before please ? really need that on my hand

If we have no specific reason for it's usage I can just use #public_send instead

seuros:
Sure thinh

ngouy:
@seuros we should be good, minor version bump and changelog updated included

seuros:
Awesome work

ngouy:
@seuros when do you think will you release it (no pressure just wondering in terms of planning)

seuros:
Tomorrow max

PR #1038, opened 2021-06-17
fix non rollbackable tenant migration

ngouy:
If you remove a column, the indexes are by default also removed
Thus this migration causes following if we try to rollback
image

2 possible solutions:
remove the `remove_index` or move it up

I chose the latter so up and down are symmetric, it's more esthetic and explicit.


seuros:
Thanks

PR #1049, opened 2021-09-30
avoid select(all_fields) in any tag query to fix error when use as subquery

kikyous:


kikyous:
@enomotodev @donquxiote someone can review this pr?

donquxiote:
@kikyous I'm not a maintainer, you probably want @seuros

seuros:
@kikyous the PR look fine, but can you add a test case ?

kikyous:
> @kikyous the PR look fine, but can you add a test case ?

ok, I will try to add a test case.

kikyous:
@seuros test case added

kikyous:
@seuros ping

andyundso:
In the meantime, if any one stumbles upon this, you can remove the `select` from the query that the gem adds with `unscope`.

```ruby
class Stuff < ApplicationRecord
  acts_as_taggable

  has_many :things
end

class Thing < ApplicationRecord
  belongs_to :stuff
end

Thing.where(stuff: Stuff.tagged_with(["foo", "bar"], any: true).unscope(:select))
```

PR #1053, opened 2021-10-19
Add support for a list of taggable IDs in the tagging conditions

moliver-hemasystems:
In the event of trying to fetch tags for multiple records, where the IDs of those records are already known, adding support for the `:id` option to accept an array of IDs to use directly, instead of using a subquery to find the same list of IDs.

I had run into a situation, where I was trying to get the tags for a series of records associated to the ones I was listing, and SQL resulting from calling `tag_counts_on` was using a subquery to fetch the IDs that were already being used in the query chain.

For example, say I was listing cases for multiple projects, and wanted to be able to filter those cases by tags on the projects.  If I did:

```ruby
Project.where(id: cases.distinct.pluck(:project_id)).tag_counts_on(:tags)
```

... it would result in a SQL that used the following sub-query:

```sql
AND taggings.taggable_id IN (
  SELECT "projects"."id"
  FROM "projects"
  WHERE "projects"."id" IN ({list of IDs})
)
```

This is causing the database to do extra work.  There's an option that can be used to pass in an ID to `tag_counts_on`, but it currently only works with a single ID.  This PR aims to extend that option to support an array of IDs as well, so the following could be used:

```ruby
Project.tag_counts_on(:tags, id: cases.distinct.pluck(:project_id))
```

... and have the generated SQL be this instead:

```sql
AND taggings.taggable_id IN ({list of IDs})
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@ Each change should fall into categories that would affect whether the release is
 
 As such, _Breaking Changes_ are major. _Features_ would map to either major or minor. _Fixes_, _Performance_, and _Misc_ are either minor or patch, the difference being kind of fuzzy for the purposes of history. Adding _Documentation_ (including tests) would be patch level.
 
-### [v8.2.0) / 2021-10-19](https://github.com/mbleigh/acts-as-taggable-on/compare/v8.1.0...v8.2.0)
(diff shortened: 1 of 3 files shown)


seuros:
@moliver-hemasystems Thank you for your contribution, can you rebase your PR ? I will release it in version 9.0.0

moliver-hemasystems:
I rebased the PR.  I didn't think of it at the time, but did you want me to adjust the changelog/version?

seuros:
Yes, please.

PR #1079, opened 2022-02-28
Adds support for horizontally sharded database projects

donquxiote:
In our project we're using a horizontally sharded database in Rails 6.1. As part of that we have to define what shards our different models can connect to through ApplicationRecord. As the gem is currently, the taggings and tags tables we use from the gem would not have access to the right connection information since they inherited from ActiveRecord::Base. It seemed like this might help more than just my project.

This PR makes the class `ActsAsTaggableOn::Tags` and `ActsAsTaggableOn::Tagging` inherit from a configurable `base_class` variable on ActsAsTaggableOn. A rake task `acts_as_taggable_on:sharded_db:install_initializer` is provided that will add an example initializer to a project.

The class passed to `base_class` is checked to be a Class object, otherwise an error is raised. 

Tests verify that `ActsAsTaggableOn::Tags` and `ActsAsTaggableOn::Tagging` inherit from ActiveRecord::Base by default or the passed class if present.


seuros:
can you rebase this PR ?

donquxiote:
@seuros more than happy to.

PR #1080, opened 2022-03-28
Implicit block expectation syntax is deprecated

kuahyeow:
Fix to pass a block instead.

Fixes errors like:

```
  1) acts_as_tagger #tag when called with a non-existent tag context should throw an exception when the default is over-ridden
     Failure/Error:
       expect(-> {
         @tagger.tag(@taggable, with: 'this, and, that', on: :foo_boo, force: false)
       }).to raise_error(RuntimeError)
     
       The implicit block expectation syntax is deprecated, you should pass a block rather than an argument to `expect` to use the provided block expectation matcher or the matcher must implement `supports_value_expectations?`. e.g  `expect { value }.to raise RuntimeError` not `expect(value).to raise RuntimeError`
```


kuahyeow:
@seuros Are you available to look at this PR ? It fixes failing specs on `master`.

Issue #1068, opened 2021-12-13
`self.find_or_create_all_with_like_by_name` does not respect many databases nor nested transactions


ayufan:
The `self.find_or_create_all_with_like_by_name` seems to implement something that resembles concurrency safe creation of records: if many records are created concurrently and one of them ends-up being unique transaction is being reversed.

This implementation is sketchy as it:

1. Does not adhere to materialised nested transactions that might use savepoints breaking transaction boundaries
2. Does not adhere to the connection used by the model

## 1. materialised nested transactions

The code can run in many level of transactions which will make `ROLLBACK` to revert the top-most which is incorrect beaviour.

```ruby
transaction(joinable: false) do # top-level
  transaction(joinable: false) do # savepoint
    ActsAsTaggableOn::Tag.find_or_create_all_with_like_by_name('AWESOME')
    # the ROLLBACK will revert to top-level and break `savepoint`
  end
end
```

## 2. Does not adhere to the connection used by the model

The Rails 6 allows to support many databases, and allows to instruct indivdual models to use a different database. The usage of `ActiveRecord::Base.` is then deemed no longer safe as it might be pointing to wrong database, and rollback might be executed on a wrong database.

```ruby
class MyModel
  connects_to database: { reading: :another_db, writing: :another_db }
end

ActsAsTaggableOn::Tag.connection_specification_name = MyModel.connection_specification_name
```

The `ActsAsTaggableOn::Tag.connection` will be different than `ActiveRecord::Base.connection`

Ref.: https://guides.rubyonrails.org/active_record_multiple_databases.html

ayufan:
Likely the proper way to implement this would be somewhere along those lines.

```ruby
...

      list.map do |tag_name|
        comparable_tag_name = comparable_name(tag_name)

        begin
          tries ||= 3
          existing_tag = existing_tags.find { |tag| comparable_name(tag.name) == comparable_tag_name }
          next existing_tag if existing_tag

          transaction { create(name: tag_name) }
        rescue ActiveRecord::RecordNotUnique
          if (tries -= 1).positive?
            existing_tags = named_any(list)
            retry
          end

          raise DuplicateTagError.new("'#{tag_name}' has already been taken")
        end
      end
    end

 ...
```

seuros:
Believed fixed on current versions. If you can still reproduce it on v13, open a fresh issue and we will take a look.

Issue #948, opened 2019-03-11
Don't issue raw rollback


0x0badc0de:
PR https://github.com/mbleigh/acts-as-taggable-on/pull/809 includes a code which issues raw rollback into a database rolling back all the changes made before calling this code. There was very valid comment regarding this https://github.com/mbleigh/acts-as-taggable-on/pull/809#issuecomment-369025242
Can this code be refactored at least to use nested transaction, raise ActiveRecord::Rollback or even rethrow original exception probably configured w/ some option. It would be much better to have an exception in caller code rather than all the changes silently discarded by raw rollback.

seuros:
Believed fixed on current versions. If you can still reproduce it on v13, open a fresh issue and we will take a look.

PR #1081, opened 2022-03-30
Handle RecordNotUnique error in find_or_create_all_with_like_by_name correctly

kuahyeow:
Related to https://github.com/mbleigh/acts-as-taggable-on/issues/1068, https://github.com/mbleigh/acts-as-taggable-on/issues/948

In the find_or_create_all_with_like_by_name method,
instead of issuing a ROLLBACK statement, we wrap the tag creation in a
transaction instead.

Issuing a ROLLBACK will break nested transactions, and also does not
respect multiple databases (available since Rails 6).

Adds a spec which attempts to simulate concurrent tag creation

### DIFF
--- a/lib/acts_as_taggable_on/tag.rb
+++ b/lib/acts_as_taggable_on/tag.rb
@@ -84,10 +84,11 @@ module ActsAsTaggableOn
         tries ||= 3
         comparable_tag_name = comparable_name(tag_name)
         existing_tag = existing_tags.find { |tag| comparable_name(tag.name) == comparable_tag_name }
-        existing_tag || create(name: tag_name)
+        next existing_tag if existing_tag
+
+        transaction(requires_new: true) { create(name: tag_name) }
       rescue ActiveRecord::RecordNotUnique
         if (tries -= 1).positive?
-          ActiveRecord::Base.connection.execute 'ROLLBACK'
           existing_tags = named_any(list)
           retry
         end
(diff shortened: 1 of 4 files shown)


kuahyeow:
Note this includes a commit from https://github.com/mbleigh/acts-as-taggable-on/pull/1080

kuahyeow:
This is now ready for a look !

2called-chaos:
Could we get the ball rolling on this please @mbleigh @seuros ? When dealing with concurrency or bulk editing this is highly annoying and frankly dangerous as the models are seemingly saved (update/update!/save/save!) when in actuality the data has been rolled back. The taggings however get created. To add to the misery the paper trail audit record gets rolled back too. Thanks

jakswa:
This is what would fix our data loss I believe as well. Thank you for trying to address it. 👍

seuros (approved):

seuros:
@jakswa can you try master directly ? if you confirm that it indeed resolve your issue, i will release a a new version.

I'm going to clean up the code base.

jakswa:
@seuros we have been testing it today and so far so good 🤞 usually we'd have a few cases per day crop up

jakswa:
Still no cases cropping back up 🎉. I'll remember to post back in here if we find any problems with this changeset, but going into the weekend feeling good! Thank you again!

PR #1082, opened 2022-04-08
Allow prefix and suffix wildcard tag search

glampr:
Improve on `wild: true` option for wildcard tag searches to also allow:
- placing the wildcard only at the suffix with `wild: :suffix` to search for SQL `awesome%`
  - (would be called prefix search, wildcard is the suffix, hence `wild: :suffix`)
- placing the wildcard only at the prefix with `wild: :prefix` to search for SQL `%awesome`
  - (would be called suffix search, wildcard is the prefix, hence `wild: :prefix`)

Let me know what you think, thanks!


seuros:
Can you rebase this PR please.

glampr:
I have rebased the branch to latest master.

seuros (approved):

PR #1101, opened 2023-03-03
Use eq instead of matches for non-wildcard searches

flickgradley:
**Context:**

Thanks for such an excellent gem! We have over 60,000 tags in our database, and optimizing the LIKE and ILIKE operators is proving to be difficult. It seems like we don't need to use LIKE/ILIKE when doing exact tag matches (non-wildcard) - this PR changes the `matches` and `matches_any` operators to `eq` and `eq_any`, which would let us use a BTREE index on lower(name) to optimize these queries. We already use lower(name) when performing a case-insensitive search, so we don't need to use ILIKE as well.


ragingdave:
@mbleigh any progress on getting this into a new release? It would be hugely helpful

PR #1102, opened 2023-03-10
Add tenant: option to tagged_with

flickgradley:
To improve multi tenancy support, I thought it would be good to add a `tenant:` option to the `tagged_with` query. We have 1.2M tagged objects in our database, and adding this tenant filter to that query should make it significantly faster to find objects tagged within a certain tenant.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ As such, _Breaking Changes_ are major. _Features_ would map to either major or m
 * Features
   * [@glampr Add support for prefix and suffix searches alongside previously supported containment (wildcard) searches](https://github.com/mbleigh/acts-as-taggable-on/pull/1082)
   * [@donquxiote Add support for horizontally sharded databases](https://github.com/mbleigh/acts-as-taggable-on/pull/1079)
+  * [@flickgradley Add tenant: option to tagged_with](https://github.com/mbleigh/acts-as-taggable-on/pull/1102)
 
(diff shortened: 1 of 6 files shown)


courtsimas:
@mbleigh can we get this added in? Would be a huge help for people.

tvongaza:
This is a great change. We are heavy users of the amazing ActsAsTenant gem and created the following rails initializer to scope our tags & taggings to the set tenant. This PRs change is a lot more flexible, though will require being a bit more explicit on the tenant setting (vs using ActsAsTenant & the require_tenant = true flag).

```
module ActsAsTaggableOnWithActsAsTenant
  def self.included(base)
    base.class_eval do
      acts_as_tenant(:account)
    end
  end
end

ActsAsTaggableOn::Tag.include(ActsAsTaggableOnWithActsAsTenant)
ActsAsTaggableOn::Tagging.include(ActsAsTaggableOnWithActsAsTenant)

module ActsAsTaggableOnTagNameValidationWithActsAsTenant
  def self.included(base)
    base.class_eval do
      # https://github.com/mbleigh/acts-as-taggable-on/blob/master/lib/acts_as_taggable_on/tag.rb#L17
      def validates_name_uniqueness?
        false
      end
      validates_uniqueness_to_tenant :name
    end
  end
end

ActsAsTaggableOn::Tag.include(ActsAsTaggableOnTagNameValidationWithActsAsTenant)
```

ragingdave:
@mbleigh any progress on getting this into a new release? It would be hugely helpful

seuros:
I will try to get some time soon to update this gem.

I avoided adding new features unless they are opt-in.

PR #1110, opened 2023-10-05
Add support to Rails 7.1

aovertus:
Update `activerecord` requirements to support Rails 7.1

PR #1114, opened 2023-10-14
Prepare release V10 release

seuros:
In this version, we will drop support for unsupported versions of rubies and activerecord.

This will allow to remove any backward compatibility hack that was introduced in the code base.

### DIFF
--- a/acts-as-taggable-on.gemspec
+++ b/acts-as-taggable-on.gemspec
@@ -15,7 +15,7 @@ Gem::Specification.new do |gem|
   gem.files         = `git ls-files`.split($/)
   gem.test_files    = gem.files.grep(%r{^spec/})
   gem.require_paths = ['lib']
-  gem.required_ruby_version     = '>= 2.5.0'
+  gem.required_ruby_version     = '>= 2.7.0'
 
   if File.exist?('UPGRADING.md')
     gem.post_install_message = File.read('UPGRADING.md')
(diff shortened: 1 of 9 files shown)


acts-as-taggable-on.gemspec
```diff
@@ -22,7 +21,7 @@ Gem::Specification.new do |gem|
     gem.post_install_message = File.read('UPGRADING.md')
   end
 
-  gem.add_runtime_dependency 'activerecord', '>= 6.0', '< 8'
+  gem.add_runtime_dependency 'activerecord', '>= 6.1', '< 7.2'
```
rjherrera, line 24: hi there @seuros! Just out of curiosity, why did you end up changing the `< 8` to `< 7.2`?. 

Reading #1110, it seemed to me that the goal was to avoid needing a new release every time a new AR minor was released but I'm not sure if you see any other risks that could come from other "non major" releases.

Thanks!

rjherrera: makes sense! thanks for the answer (and the release!)

Skulli: as predicted https://github.com/mbleigh/acts-as-taggable-on/issues/1130 Rails 7.2 released

seuros: Let release a new version.

PR #1142, opened 2024-12-12
docs: Add a section about uuid primary key in README.md

elthariel:
It's becoming frequent to use UUIDs as Primary Key in applications, and it happens that one forgets to update the generated migration. In the case of acts-as-taggable-on, it leads to a very confusing error "Tag has already been taken".

This PR adds some note about the subject in the README


seuros:
This is complicated, because this gem only support one type of ids globally.

I build another gem https://rubygems.org/gems/no_fly_list https://github.com/contriboss/no_fly_list , that handle this case.

PR #1154, opened 2025-06-09
Fix issue when reordering tags and @@remove_unused_tags is true

james-reading:
When reordering taggings, the taggings get deleted and  recreated. However, by the time the taggings are recreated, the tags have already been removed.

Fixed by wrapping the delete + create in a transaction, and only removing unused tags on transaction commit.


seuros:
Can you rebase on master ? i'm hesitant to merge it  as people using pgbouncer, could have transaction disabled.

seuros:
Active Record allow you to check if the connection support transactions.

Can you check if the versions  are raising or just ignoring it.

PS: if only 7.2 is problematic, we can drop support. But if they all raise , we will need to find a proper solution.

PGbouncer disable transaction by default if i recall correctly.

PR #1175, opened 2026-04-17
Performance improvements

fmichaut-diff:
Re-opening https://github.com/mbleigh/acts-as-taggable-on/pull/1121 

For some reason, I did not receive any notifications about the comments / pings, sorry about that

Original Description: 

----

Opening this pull request cause we noticed that this gem was making duplicate queries for the same resources on our app, and decided to dig in to find the root cause.

The main usecases were : 
- On delete, we notice that we the gem delete a Tagging, and then query the same Tag twice
- On `save!` with nothing changed, the gem queries existing Tags and Tags in current context, even tho nothing changed so the gem should not have to query anything
- On `save!` with changes (or not), the gem seems to query the same contexts multiple times, which is redundant

Here are improvements for the first and last usecase. The middle one seem more complex, and is left for a future improvement.

# Improvements 
For the delete improvement : 
We simply only reload the `tag` record if already loaded. Otherwise `tag.reload` would load it twice.

For the `save!` improvement : 
Since we initialize `custom_contexts` with `taggings.map(&:context).uniq`, but `tagging_contexts` uses both `self.class.tag_types.map(&:to_s)` and `custom_contexts`, we have duplicates contexts in `tagging_contexts` which results in the gem attempting to save each context twice.
Add a `uniq` to `tagging_contexts` to ensure that does not happen.

### DIFF
--- a/spec/acts_as_taggable_on/tagging_spec.rb
+++ b/spec/acts_as_taggable_on/tagging_spec.rb
@@ -64,6 +64,33 @@ RSpec.describe ActsAsTaggableOn::Tagging do
     ActsAsTaggableOn.tags_counter = tags_counter_previous_setting
   end
 
+  context 'when removing unused tags' do
+    before do
+      ActsAsTaggableOn.remove_unused_tags = true
+    end
+
+    it 'should call reload on tag if it was loaded' do
+      @taggable = TaggableModel.create(name: 'Bob Jones')
+      @taggable.tag_list.add('aaa')
+      @taggable.save
+      tagging = @taggable.taggings.first
+      tagging.tag # load the tag
+
+      expect(tagging.tag).to receive(:reload).once
+      tagging.destroy
+    end
+
+    it 'should not call reload on tag if it was not loaded' do
+      @taggable = TaggableModel.create(name: 'Bob Jones')
+      @taggable.tag_list.add('aaa')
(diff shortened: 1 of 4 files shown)


fmichaut-diff:
@n-rodriguez if you have time to review this it'd be great

n-rodriguez:
@fmichaut-diff LGTM. A benchmark or a test on the issued queries would be much appreciated 👍

fmichaut-diff:
Before : 
image

After :
image

Added a spec for it as well

fmichaut-diff:
Anything else I need to do on this ?

fmichaut-diff:
@n-rodriguez ?

n-rodriguez:
LGTM ping @seuros thank you!

spec/acts_as_taggable_on/tagging_spec.rb
```diff
@@ -64,6 +64,33 @@
     ActsAsTaggableOn.tags_counter = tags_counter_previous_setting
   end
 
+  context 'when removing unused tags' do
+    before do
+      ActsAsTaggableOn.remove_unused_tags = true
```
seuros, line 70: this need to be set to it initial value  after each test.  

use after or around block.

seuros (changes requested):

PR #1178, opened 2026-07-22
Use lease_connection to avoid ActiveRecord::Base.connection deprecation

benaitcheson:
Rails 7.2 soft-deprecated the model-level `ActiveRecord::Base.connection` in favour of `lease_connection` ([rails/rails#51230](https://github.com/rails/rails/pull/51230)). `ActsAsTaggableOn::Utils.connection` calls `ActsAsTaggableOn::Tag.connection`, so every tag operation that goes through `Utils` (e.g. `using_postgresql?`, `like_operator`) emits:

```
DEPRECATION WARNING: Called deprecated `ActiveRecord::Base.connection` method.
Either use `with_connection` or `lease_connection`.
 (called from ActsAsTaggableOn::Utils.connection at .../acts-as-taggable-on/utils.rb:10)
```

This switches `Utils.connection` to `lease_connection` when the model responds to it, falling back to `connection` for ActiveRecord < 7.2 (the gem still supports `>= 7.1`). `lease_connection` returns the same leased-per-thread connection that `connection` did, so behaviour is unchanged.

Added specs covering both the `lease_connection` path and the legacy fallback.

### DIFF
--- a/lib/acts-as-taggable-on/utils.rb
+++ b/lib/acts-as-taggable-on/utils.rb
@@ -6,8 +6,16 @@ module ActsAsTaggableOn
   module Utils
     class << self
       # Use ActsAsTaggableOn::Tag connection
+      #
+      # Rails 7.2 deprecated the model-level `.connection` in favour of
+      # `lease_connection` (rails/rails#51230). Prefer it when available and
+      # fall back to `connection` for ActiveRecord < 7.2.
       def connection
-        ActsAsTaggableOn::Tag.connection
+        if ActsAsTaggableOn::Tag.respond_to?(:lease_connection)
+          ActsAsTaggableOn::Tag.lease_connection
+        else
+          ActsAsTaggableOn::Tag.connection
+        end
       end
 
       def using_postgresql?
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,9 @@ As such, _Breaking Changes_ are major. _Features_ would map to either major or m
 
 ### unreleased
 
+* Fixes
+  * Use `lease_connection` in `ActsAsTaggableOn::Utils.connection` to avoid the Rails 7.2 `ActiveRecord::Base.connection` deprecation warning, falling back to `connection` on ActiveRecord < 7.2
+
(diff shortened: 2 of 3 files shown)


benaitcheson:
@seuros if you have a moment, could I please get a review on this?

lib/acts-as-taggable-on/utils.rb
```diff
@@ -6,8 +6,16 @@ module ActsAsTaggableOn
   module Utils
     class << self
       # Use ActsAsTaggableOn::Tag connection
+      #
+      # Rails 7.2 deprecated the model-level `.connection` in favour of
+      # `lease_connection` (rails/rails#51230). Prefer it when available and
+      # fall back to `connection` for ActiveRecord < 7.2.
       def connection
-        ActsAsTaggableOn::Tag.connection
+        if ActsAsTaggableOn::Tag.respond_to?(:lease_connection)
```
seuros: why the defensive coding ?

benaitcheson: Are you questioning the conditional or me updating a module marked as "to be deleted"?

Gemspec has >= 7.1 which is why the lease_connection check.

seuros: Oh right ! Let drop support for 7.1 and 7.2. 
I got confused with [NFL](https://github.com/contriboss/no_fly_list) which is a drop-in replacement of AATO.

8.2.0.alpha is current, and bring more breaking changes.

benaitcheson: Addressed here: [`b915ca8` (this PR)](https://github.com/mbleigh/acts-as-taggable-on/pull/1178/commits/b915ca85dd63256986f8437e707de8409adfff2c)

Happy to split to a separate PR if you'd prefer

seuros:
Its fine! The commits are clear. We could drop Rails 7.2 as well (people must upgrade)

No need to add a test. A test like this doesn’t really provide any value; it's basically an assert true. If that line were wrong, the entire gem would stop working anyway.

Ill merge this and cut a release tomorrow morning (Friday).

Thank you for the help.

